Issue #24, opened 2025-07-23
Draw areas


LanesGood:
Allow users to draw areas.

Area draw tools should allow for:
- polygon
- delete

When creating a new area, users should be required to confirm their area. Multipolygon areas are permitted.

Once confirmed, the area geojson will be ready for upload.

Area tools will have a cancel/confirm button interface included.

When drawing, the following should be disabled:
- chat
- other map area tools
- interaction with other map areas (selection, etc)

To do list:

- [x] Upload to the API (when ready) as a feature collection
- [ ] Disable other parts of the UI
- [ ] Delete areas (maybe in another ticket?)
- [x] Check area of drawn area, ensure under maximum size
- [x] Style drawn polygon differently based on area (red over max size)
- [x] Ensure crosshair cursor is reset to initial after leaving drawing mode

vgeorge:
In https://github.com/wri/project-zeno-next/pull/41 we added the draw feature.

@geohacker @batpad suggested that we change the generate feature from MultiPolygon to a FeatureCollection. 

@LanesGood I added a todo list to your comment to track the remaining work.

vgeorge:
@LanesGood that would be more involved because we would need to enable polygon editing. I suggest we treat this as an enhancement in a separate ticket.

About disabling other parts of the UI, which user actions do we want to block? I'm wondering if we can apply an overlay to the side panel. Otherwise we would have to add a disabled state to each child, which also seems fairly complex. Should we open a separate ticket for this as well?

I can take a look at the crosshair fix.

LanesGood:
Closing, enhancements ticketed

PR #64, opened 2025-07-24
Limit custom areas

vgeorge:
Contributes to #24.

This adds minimum and maximum area sizes for area upload and draw. The current limits are set to 0.1 - 10,000 km².

For area drawing, the validation error is only displayed when the user tries to confirm, and the area is cleared if it's invalid. This isn’t the ideal workflow, but it's much simpler than validating area size during the drawing process and will still prevent users from submitting invalid areas. We can improve the UX later when possible.

### Large area draw
draw-large-area

### Small area draw
draw-small-area

### Large area upload
upload-large-area

This is ready for review. I suggest to test all possible combinations (too small, ok, too large) in draw and upload modes.

### DIFF
--- a/app/components/MapAreaControls.tsx
+++ b/app/components/MapAreaControls.tsx
@@ -20,6 +20,7 @@ import { LayerId, selectLayerOptions } from "../types/map";
 import useMapStore from "../store/mapStore";
 import { Tooltip } from "./ui/tooltip";
 import { MAX_AREA_KM2, MIN_AREA_KM2 } from "../constants/custom-areas";
+import { formatAreaWithUnits } from "../utils/formatArea";
 
 function Wrapper({
   children,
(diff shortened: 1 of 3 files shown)


app/components/MapAreaControls.tsx
```diff
@@ -185,6 +187,31 @@ function MapAreaControls() {
...
+          color="red.700"
+          fontSize="sm"
+        >
+          <Box fontWeight="bold" mb={1}>
+            {validationError.code === "too-small"
+              ? "Area is too small"
+              : "Area is too large"}
+          </Box>
+          <Box fontSize="xs">
+            Your area: {validationError.area.toLocaleString()} km²
```
LanesGood, line 208:
```suggestion
            Your area: {validationError.area.toLocaleString().toFixed(2)} km²
```

@vgeorge can we limit the decimals here?

app/components/MapAreaControls.tsx
```diff
@@ -185,6 +187,31 @@ function MapAreaControls() {
...
         </Box>
       )}
+      {validationError && (
+        <Box
+          px={3}
+          py={2}
+          bg="red.50"
+          borderColor="red.200"
+          borderWidth="1px"
+          borderRadius="md"
```
LanesGood, line 197:
```suggestion
          borderRadius="sm"
```

Issue #21, opened 2025-07-22
Connect New Conversation actions


LanesGood:
The "New Conversation" action button is stubbed on the frontend and needs to be connected to enable:
- new blank conversation
- new conversation from template

<img width="183" height="128" alt="Image" src="https://github.com/user-attachments/assets/23fc2cc6-0e55-4756-9d25-9a8632c91073" />

PR #65, opened 2025-07-25
Remove new conversation options

LanesGood:
Removes menu dropdown and option for "new conversation from template" for MVP.

NB: We can revert this when conversation templates are ready.

Contributes to #21


vgeorge (approved):
@LanesGood 👍

PR #68, opened 2025-07-25
Map area tag style changes

LanesGood:
Uses the native Chakra UI `Tag` component for styling. This reduces the need for multiple components and also tightens up the styling some.

This also reduces the need for style props and passing mode-dependent colors.

Some noted changes:
- Selected color is currently not 100% matching design but this is dependent on a change to the theme.
- I've added the "Remove" `X` button in at all times to avoid layout shifts; just reducing opacity when unhovered. Updated in design as well

image
image


kamicut (approved):
LGTM!

PR #74, opened 2025-07-25
Fix missing upload dialog

LanesGood:
This pr:
- fixes the missing upload dialog (turned out we were missing this on the `thread/[id]` page
    - I'm not 100% sure why this page is needed vs. having a single page, and adding the id if available vs. new
- componentizes the header to use on the `thread/[id]` page


vgeorge (approved):
@LanesGood good catch. I did a refactor to reuse the UI in different routes. It would be nice if you could take a look before merging.

Issue #27, opened 2025-07-23
Custom Area Management


LanesGood:
We will support adding custom areas by the user — draw, upload, and manage them on the frontend for the user to add/remove from their context.

yellowcap:
Seems mostly done, but the nice edit features are missing and we need to decide on the areas panel

@01painadam is it ok if we can't delete custom areas for P0? (has implications for past thread rendering)

@LanesGood what is the plan for the areas panel in the chat window now that we manage mostly throug the map layer?

LanesGood:
@yellowcap re: areas context menu - in #32, I captured comments from discussion with @kamicut. We still want to make this area context menu available as it allows a text/db search experience for the custom areas, rather than just a spatial search. However we do now have spatial search available, so as mentioned in that ticket we can belay this if absolutely necessary.

PR #92, opened 2025-07-31
feat: connect custom areas to the API

vgeorge:
Contributes to #27.

~~**Work in progress.**~~

Changes introduced:

* **API proxy route:** The app authentication workflow is not set up to make direct requests to the backend API via client components. This adds a generic route, `/api/proxy/[...path]`, which forwards requests to the backend while applying the correct auth headers. This logic already exists in other routes, and we should consolidate it when possible
* **TanStack Query:** Introduced to reduce code repetition, improve type safety, handle mutations, and provide a more robust approach compared to using `fetch` directly
* **Zod:** Added to ensure type-safe responses from the API by defining schemas for both request and response payloads
* **useCustomAreas hook**: to handle area creation and fetching

At the moment, this just attempts to send a mock GeoJSON to the API when drawing finishes. I haven't been able to get it working yet because the staging API appears to be down. I plan to set up the backend API locally to continue development.

Potential next steps:

* Test with local backend API
* Connect `POST /api/custom-area`
* Connect `GET /api/custom-area` to display areas from the API on the map

This might already be a large enough scope, we can handle custom area selection in a separate PR.

cc @kamicut @LanesGood

### DIFF
--- a/app/api/proxy/[...path]/route.ts
+++ b/app/api/proxy/[...path]/route.ts
@@ -3,50 +3,87 @@ import { NextRequest, NextResponse } from "next/server";
 
 const TOKEN_NAME = "auth_token";
 const BASE_URL = "https://api.zeno-staging.ds.io/api";
+const METHODS_WITH_BODY = ["POST", "PUT"];
 
+async function getAuthToken(): Promise<string | null> {
+  const cookieStore = await cookies();
+  return cookieStore.get(TOKEN_NAME)?.value || null;
+}
+
+/**
+ * Builds the target URL for the proxy request. The upstream API expects
+ * a trailing slash.
+ * @param pathSegments - The path segments to append to the base URL.
+ * @returns The target URL with the path segments.
+ */
+function buildTargetUrl(pathSegments: string[]): string {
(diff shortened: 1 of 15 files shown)


vgeorge:
This is ready for review.

Suggested steps:

- Draw area and confirm area, reload the page, draw areas are displayed
- Upload area from Geojson file, reload the page, uploaded areas are displayed

This PR covers the existing tools realted to custom areas. 

Area deletion, enable/disabling custom areas layers would be implemented in separated PRs.

batpad:
> API proxy route: The app authentication workflow is not set up to make direct requests to the backend API via client components. This adds a generic route, /api/proxy/[...path], which forwards requests to the backend while applying the correct auth headers. This logic already exists in other routes, and we should consolidate it when possible

@vgeorge I don't understand why we need to do this? Can't this just be some helper on the client-side to add the correct auth headers, etc. and just make the request from the browser directly to the API? I'm just a bit concerned of adding a proxy redirection here, which would also mean that all requests to the backend would now come from the NextJS backend component and not from user's browsers? Am guessing it's **okay**, it just feels strange to me to add a layer of backend proxy here.

PR #71, opened 2025-07-25
Welcome Modal

wrynearson:
Closes https://github.com/wri/project-zeno-next/issues/59

image

This builds a welcome modal component.
- Input is focused on the prompt box
- Prompts are stored in public/welcome-prompts.json

To handle separately:
- allow users to toggle "don't show this again"
- remove other welcome message in chat
- handle potential modal conflict with authentication modal (likely by having this auth flow triggered in a different way)

PR #101, opened 2025-08-06
Update welcome modal

LanesGood:
Makes the changes suggested in #71 and also updates styles.
- fixes click handlers
- adds prompt input pattern from chat message
- updates styles
- updates prompts


wrynearson (approved):

LanesGood:
@wrynearson I merged develop into your branch, and then I rebased off your branch - looks like this is the right approach!

PR #103, opened 2025-08-06
Implement date picker to add date to context

danielfdsilva:


LanesGood (approved):
This looks great to me! I see the following when I select a date (single day range test):
image

I do think that it would be nice to _only_ show the date resolution selected, but think this is up to opinion and likely an enhancement. That is to say, if a user selects "Month" resolution, the context chip/tag would simply state "March 2009 - April 2010" rather than including the day values.

app/components/ContextMenu.tsx
```diff
@@ -372,3 +261,141 @@ function ContextMenu({
...
+      )?.id;
+      if (ctxId) {
+        contextStore.removeContext(ctxId);
+      }
+
+      contextStore.addContext({
+        contextType: "date",
+        content: `${format(dateValue[0], "yyyy-MM-dd")} - ${format(
+          dateValue[1],
+          "yyyy-MM-dd"
```
LanesGood, line 348:
```suggestion
        content: `${format(dateValue[0], "yyyy-MM-dd")} — ${format(
          dateValue[1],
          "yyyy-MM-dd"
```
Use emdash to separate date start and end.

Also, I imagine we could use the `view` option from the state to change the format function here if we wanted to update the displayed date based on the selected date resolution (per my review comment). We'd want to ensure this affects only the display and not the stored context value.

vgeorge:
I've fixed conflicts and will merge. @LanesGood, I created a ticket with suggestions in https://github.com/wri/project-zeno-next/issues/110.

PR #107, opened 2025-08-07
Logout from RW using window popup

### LINKED ISSUE #84: Non-Whitelist authentication bug
When someone that has tried authentication is not on the whitelist, the frontend stores a cookie for the WRI login but blocks the frontend for that user. This prevents the user from trying to log in with a different email.

kamicut:
Fixes #84 by logging out from RW. I used the same postMessage approach as logging in to close the popup on "done". There is a limitation here because after logging out, the `callbackUrl` is not calling our callback function, so I set a timeout to close the popup. Once we have a proper redirect from the authentication proxy then it will auto-close.

### DIFF
--- a/app/components/LoginOverlay.tsx
+++ b/app/components/LoginOverlay.tsx
@@ -137,7 +137,6 @@ function LoginOverlay() {
         "WRI Logout",
         "width=600,height=700,scrollbars=yes,resizable=yes,status=no,location=no,toolbar=no,menubar=no"
       );
-      console.log("Popup:", popup);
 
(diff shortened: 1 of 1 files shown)


vgeorge (changes requested):
@kamicut I'm not able to test this change. I'm currently logged into staging as `vitor@devseed`, and I don't see a way to log out and switch to `vitor@gmail`. Even after clearing browser data, when I click "Login with WRI", the popup skips account selection and logs me in directly with `vitor@devseed`.

PR #114, opened 2025-08-08
Dataset tile urls

kamicut:
- [x] Adds ability to display datasets from pick-dataset tool
- [x] Adds the dataset to context
- [x] Styling
- [ ] Legend
- [x] Context Menu
- [x] Connecting to Chat LLM

This adds the ability to display datasets like "Tree Cover Loss" on the map. I'm not too happy with the styling of the dataset card, so I could use some help there @LanesGood. Also, I'm not sure what the legend should be, and what data should be returned in the dataset object for that.

For the context menu, I need to show that the current data layer is a selection from a list of tile urls that are hard coded in the menu. For that I'll need the list of definite tile layers that we can use for analysis @yellowcap. For the LLM, I also need to know the ui_context key to use for communication.

image


vgeorge:
@kamicut should I review this now? what are the steps to test?

kamicut:
@vgeorge the functionality is now working -

1. Through the chat, you can ask for "Tree Cover Loss in Lebanon" and see that the context is correctly assigned (and also selected in the context menu). The chat bar will also display a button to add the dataset to the map.
2. Through the context menu, select a dataset and then ask something in the chat like "I'm interested in Lebanon" and see that the LLM correctly understands the context.

app/store/chatStore.ts
```diff
@@ -187,6 +169,14 @@ const useChatStore = create<ChatState & ChatActions>((set, get) => ({
       };
     }
 
+    const datasetContext = context.find((ctx) => ctx.contextType === "layer");
+    if (datasetContext && typeof datasetContext.datasetId === "number") {
+      const ds = DATASET_BY_ID[datasetContext.datasetId];
+      if (ds) ui_context.dataset_selected = { dataset: ds };
+    }
+
+    console.log("ui_context", ui_context);
```
vgeorge, line 178: Remove

01painadam:
> How a user might want to re-add, I'm not sure

Through the manual dataset selector?

> Separately, I'm not sure what it means to "Show on map" for the other widgets, and think we need to ensure this is mapped out for frontend implementation.

The show on map interaction was originally tied to a [chart insight](https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/WRI_LCL_ProjectZeno?node-id=2330-2133&t=PrFpMtpkbMNzFu8B-4). We'd originally considered that each widget has a dataset attached to it that can be visualised. I feel like this behaviour could be moved to the chart component.

In principle I actually like how this informs the user about the dataset being selected (complete with description), but I agree that I prefer the controls to live elsewhere e.g. in the legend and chart

01painadam:
> ok - so I'll just re-use the dataset card components and make their state based on the context. How about the image? Should it be the same default map image that we have or should they be different based on the data layer?

Its subtle but they are different - they each show a snapshot of the data against the map. @faustoperez maybe we can generate some static images per dataset for these to make it more obvious? Either that or pull a tile png @yellowcap

PR #115, opened 2025-08-11
Update custom area type to be consistent with other types

sunu:
Needed to get custom areas to work in the chat as part of https://github.com/wri/project-zeno/pull/297

### DIFF
--- a/app/components/map/layers/CustomAreasLayer.tsx
+++ b/app/components/map/layers/CustomAreasLayer.tsx
@@ -29,7 +29,7 @@ function CustomAreasLayer() {
               src_id: id,
               name,
               source: "custom",
-              subtype: "custom_area",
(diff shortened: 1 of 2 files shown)


sunu:
@LanesGood In the current state, I think this PR doesn't make any difference. You should be able to get the same result that you've shown in the screenshot without the changes in this PR.

But after https://github.com/wri/project-zeno/pull/297 is merged and deployed and https://github.com/wri/project-zeno/issues/298 is fixed by the WRI team on the analytics API side, it should be possible to get valid results and insights back for custom areas using this PR for the same query that you've shared.

So for now, I think it's ok to either merge as is since this is basically a no-op on the frontend or wait for the relevant backend PRs to be merged before we test and merge this PR.

PR #117, opened 2025-08-11
Implement minor usability enhancements

01painadam:
A couple of small usability tweaks to the chat UI in order to reduce perceived wait times.

- [x] Adds typewriter effect
- [x] Adds "thinking" message while awaiting first assistant message

Feel free to bin :)

### DIFF
--- a/app/components/ThinkingMessage.tsx
+++ b/app/components/ThinkingMessage.tsx
@@ -1,0 +1,43 @@
+import { Box, Flex } from "@chakra-ui/react";
+
+export default function ThinkingMessage() {
+  return (
+    <Box
+      display="flex"
+      alignItems="center"
+      bg="whiteAlpha.200"
+      color="fg"
+      px={4}
+      py={3}
+      borderRadius="lg"
+      maxW="80%"
+      mb={4}
+    >
+      <Flex alignItems="center" gap={1}>
+        <span>Thinking</span>
+        <span className="thinking-ellipsis">
+          <span>.</span>
+          <span>.</span>
+          <span>.</span>
+        </span>
+      </Flex>
+      <style jsx>{`
+        .thinking-ellipsis span {
+          opacity: 0.2;
+          animation: blink 1.4s infinite both;
+          font-weight: bold;
+        }
+        .thinking-ellipsis span:nth-child(2) {
+          animation-delay: 0.2s;
+        }
+        .thinking-ellipsis span:nth-child(3) {
+          animation-delay: 0.4s;
+        }
+        @keyframes blink {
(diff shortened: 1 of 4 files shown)


LanesGood (approved):
@01painadam looks great! I'd explored using some various libraries for typewriter affect in [a different branch](https://github.com/wri/project-zeno-next/compare/develop...add/typewriter-effect), but think that the non-library approach is actually better given the lack of dependencies. 

A few things we may want to consider supporting in the future:
- stopping typing animation if an error occurs or the user interrupts the LLM
- adding a typing cursor/mark that follows the text

LanesGood:
One additional fix to consider: how to handle "live-typing" of past thread responses? I was facing this issue as well in my previous attempt; we should decide whether there are situations where we want this to be typed, or if the text should just render.

app/components/ThinkingMessage.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <Box
+      display="flex"
+      alignItems="center"
+      bg="whiteAlpha.200"
+      color="fg"
+      px={4}
+      py={3}
+      borderRadius="lg"
+      maxW="80%"
+      mb={4}
```
LanesGood, line 14:
```suggestion
      mb={4}
```
I think the only styling needed here is the margin; it's not a button, and there is only one child.

01painadam:
Cheers for the initial review @LanesGood - how do you feel about it from a look/and feel perspective? 

I've added todo's to address the other edge cases you raised.

app/components/TypewriterText.tsx
```diff
@@ -0,0 +1,32 @@
+import { useEffect, useState } from "react";
+
+interface TypewriterTextProps {
+  text: string;
+  speed?: number;
+  render?: (text: string) => React.ReactNode;
+  onDone?: () => void;
+}
+
+export default function TypewriterText({ text, speed = 15, render, onDone }: TypewriterTextProps) {
```
LanesGood, line 10:
```suggestion
export default function TypewriterText({ text, speed = 25, render, onDone }: TypewriterTextProps) {
```
I think a slightly faster speed feels more natural, and smoother.

LanesGood:
For reproducing the issue I'm seeing with previous message animation: in a current thread, after typing a new message, the previous message re-types itself while the response to the new message is still in the "thinking" state.

app/components/TypewriterText.tsx
01painadam: Agreed. I played with making it variable (based on message length) but it was getting overly complex...

Issue #39, opened 2025-07-23
Connect chat response actions


LanesGood:
Users should be able to execute the following actions on each chat response:
- copy response text
- Provide response feedback
    - good response
    - bad response
- Regenerate response

<img width="138" src="https://github.com/user-attachments/assets/f1cde47c-34ca-4163-a972-4b08f15bb8d5" />

01painadam:
How are 👍 / 👎  connected (if at all) to the agent? Will these be logged in traces for example? cc: @srmsoumya

LanesGood:
I think we could include a feedback modal on negative responses. This would be a three step process
1. Click "thumbs down"
2. Click on an expand button to provide feedback
<img width="25-" alt="Image" src="https://github.com/user-attachments/assets/da178b65-e37d-4a49-9f09-9b70b33ded6b" />
3. Enter text and send, or cancel

<img width="250" alt="Image" src="https://github.com/user-attachments/assets/2db47e62-e1fc-4e36-b8d1-a51182ae5c4b" />

01painadam:
@kamicut whats the status on this one?

kamicut:
@01painadam this one is done on the frontend side, but we are investigating why the responses are not showing up in Langfuse cc @leothomas @batpad

PR #129, opened 2025-08-13
Copy message response

LanesGood:
Contributes to #39.

https://github.com/user-attachments/assets/0b299a5f-a100-4251-9b12-0265846369a5


LanesGood:
Merging after discussion with @kamicut

PR #135, opened 2025-08-14
Sidebar usability enhancements

01painadam:
Adds accordion groups to the sidebar (open by default) so users can collapse groups for easier navigation.

Todos
- add kebab menu to each thread item with name, rename (and share?)

### DIFF
--- a/app/sidebar.tsx
+++ b/app/sidebar.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useMemo } from "react";
 import {
   Button,
   Circle,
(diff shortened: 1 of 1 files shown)


LanesGood:
Such a great addition, glad you took this on! I'm separately trying to figure out the grouping & timestamps; I noticed there were some issues with how threads were being grouped and traced it to the thread timestamp being off. Rename and delete from the sidebar is a huge improvement. Testing functionality now.

01painadam:
Cheers for the refactor @LanesGood - much tidier :) 

Before merging, could you take a quick look at the keyboard focus? I can't quite pin what I've changed that's messed with the original behaviour.

Before:
Screenshot 2025-08-15 at 8 26 53 AM

After:
keyboard-focus-2025-8-15

PR #137, opened 2025-08-15
Update theme values

LanesGood:
Updates theme values to the correct colors and typography. (Application font was previously not loaded correctly).

This replicates the last two commits of #105 as I suspect the review of that may take longer and this seems useful to get added sooner.


LanesGood:
cc @01painadam (it was your screenshots that made me realize fonts weren't loaded correctly)

01painadam:
Nice spot!

PR #148, opened 2025-08-18
Add typewriter text effect [ DO NOT MERGE ]

01painadam:
This PR add a typewriter effect to decrease perceived user wait times. 

New Features:
- Typewriter effect for messages in 'realtime' (read: 'live') conversations
- typewriter disabled on load of old conversations

typewriter-2025-8-18

As a consequence of adding the above, some pretty significant changes are made to messages so that the typewrite effect is disabled on load of past conversations. This will need serious review.

TL;DR - changes

1. The fetchThread function in the chatStore was refactored. Instead of adding messages to the state as they arrive, it now collects all incoming messages into a temporary array. Once the stream is complete, it performs a single, atomic update to the store, adding all messages and setting isFetchingThread to false in the same operation. This prevents multiple re-renders and ensures a clean "all-at-once" display of the conversation. **Note that this does slowdown render of messages in realtime chats. Should we consider not batching is this case?** 
2. A new optional property, source: 'historical' | 'realtime', was added to the ChatMessage type. All messages loaded via fetchThread  are now marked as 'historical'. The  MessageBubble component now uses this immutable message.source property to determine whether to skip the typewriter animation.

### DIFF
--- a/app/components/ChatMessages.tsx
+++ b/app/components/ChatMessages.tsx
@@ -9,6 +9,11 @@ function ChatMessages() {
   const containerRef = useRef<HTMLDivElement>(null);
   const { messages, isLoading, isFetchingThread } = useChatStore();
 
+  const lastAssistantIdx = messages
+  .map((msg, idx) => (msg.type === "assistant" ? idx : -1))
+  .filter(idx => idx !== -1)
+  .pop();
+
   // Auto-scroll to bottom when new messages are added
   useEffect(() => {
     if (containerRef.current) {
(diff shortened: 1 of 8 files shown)


faustoperez:
I think we have already a designed behavior [here](https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/WRI_LCL_ProjectZeno?node-id=2030-14369&t=9yg4O51X7IcbiI8M-1). 

A spinner + gradient on the "reasoning" text is a bit cleaner in my opinion.

image

I like the typewriter effect for the following response.

Another thing we could do is clear the chat view with every new prompt. The chat history is still available by scrolling up, but each prompt starts with a clean window. It's the way chatGPT works and I think it's quite neat.

PR #152, opened 2025-08-20
Amplify config

batpad:
Adds an amplify.yml that's needed for Amplify to be able to deploy the app.

cc @kamicut @danielfdsilva


batpad:
Going to go ahead and merge this as the amplify deploy seems to all work.

PR #153, opened 2025-08-20
Allow anonymous users to browse the app

danielfdsilva:
Initial code for anonymous access. Needs to be fully tested once the api is ready.

The secret api key should be set using `NEXT_SERVICE_API_KEY` env.


kamicut:
@danielfdsilva I've changed the name of the custom header to `X-ZENO-FORWARDED-FOR`

@LanesGood let's make those UI changes in a separate PR - this one is to add headers to requests for the backend to work.

kamicut (approved):

PR #157, opened 2025-08-25
Add marquee to title

01painadam:
Added bonus! Adds a ticker effect to the title on hover, avoiding the need for a tooltip.

title-ticker-2025-8-25 (1)

Did not add looping as the code started getting bloated, but this could in principle be added.

Questions:
- Is this a better solution than just the tooltip?
- Is the ticker and rewind rewind speed ok?
- Should this effect be applied inside the chat history too, or will it be too busy?

### DIFF
--- a/app/ChatPanelHeader.tsx
+++ b/app/ChatPanelHeader.tsx
@@ -1,4 +1,6 @@
-import { useCallback, useMemo } from "react";
+"use client";
+
+import { useCallback, useMemo, useRef, useState, useEffect } from "react";
 import {
   Flex,
   IconButton,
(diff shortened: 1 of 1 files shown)


faustoperez:
Looks good, my only concern is accessibility (which we'd also have if we use a tooltip).

[This article](https://www.sitelint.com/blog/truncating-text-and-making-it-accessible) provides a solution for touchscreens, which involves tapping on the ellipsis text, causing it to animate.

But that clashes with the dropdown behaviour.

ChatGPT shows the conversation text truncated on mobile and doesn't provide a solution.

Maybe we can set the LLM to provide short conversation descriptions?

faustoperez (approved):
Looks good, left a general comment about accessibility on the issue.

LanesGood (changes requested):
Suggest #158 in place of marquee.

PR #168, opened 2025-08-26
Use query params for initial prompt

### LINKED ISSUE #143: Enable query parameter for initialPrompt
Sent a initialPrompt as a query parameter - will help fix the prompt hero functionality in #75 .

LanesGood:
Uses query params to send prompt from homepage. Closes #143.


LanesGood:
NB: I'm seeing a lot of the duplicate thread messages and thread creation, but these should be fixed by #147.

Tested with non-latin characters (not sure if encoding was necessary for this, but seems to be good practice for protection nonetheless):
image
image

LanesGood:
@danielfdsilva it looks like even after merging in your changes from #151, it is necessary to change the component to `/app/thread/{id}` in order for this query param PR to work.

LanesGood:
This is now dependent on #181 being merged in first.

LanesGood:
This is working with 181 merged in, but there are still a few issues:

1. I now see the following error in CI:
`useSearchParams() should be wrapped in a suspense boundary at page "/app". Read more: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout`

I'm hesitant to introduce a <Suspense> component without knowing exactly where to do this, though in general this makes me realize that (I think) it would be good in general for the application to have an error boundary and/or fallback state.

2. I am also seeing a lot of duplication when I send the prompt from the search params via the landing page - both duplicate threads, and duplicate messages in the thread.

3. I have an intermittent error of the thread reloading/resetting after the LLM returns the response. What I see is:
    a. Send the prompt via the "Go" button on the landing page input prompt box
    b. Prompt correctly is set in query params, and sent to the LLM
    c. While prompt is in URL and LLM is returning first streamed message, the chat is displayed normally. However, after the LLM message is completed, the URL updates to the new thread, and the chat is erased. This seems related to the problems reported in #183 

    This is an intermittent issue but I'm curious if it can be reproduced. My most recent change in https://github.com/wri/project-zeno-next/pull/168/commits/007415b446d2b6b4e824675217351c04938c2d11 seems to have fixed the issue but am not 100% sure.

@danielfdsilva @kamicut would be good to get your testing and input here. I understand we may hide this feature behind a flag anyways given the updates to launch plans; but perhaps these issue is still important to address for the overall app.

01painadam:
> I've removed the hardcoded array of sample prompts - so both the prompt marquee and the sample prompt input box on the homepage are now using the same set of prompts from the `welcome-prompts.json` file. I will say, though, this loads a little slower than is ideal. cc @01painadam
Nice! I'll open a PR on optimisation - the landing page in particular could use a bit of compression.

> Lastly, we do see an error in development thanks to a 'bad setState' - updating the promptIndex in the LandingPage via an interval in the Hero section seems to be a code smell for react. We can either A) ignore this, as its not affecting production, or B) remove the functionality of clicking a scrolling prompt in the marquee to set the prompt in the input box. I think either are acceptable solutions with my preference being B - remove the functionality (which we're doing for launch anyways).
Agree. If we remove the marquee we still have the rolling prompts populating the chat input right?

LanesGood:
@01painadam will the landing page for the public/open stage still feature the prompt box? This PR is ready to go (@kamicut I can fix the merge conflicts) - I see you closed the related issue.

kamicut (approved):
Tested this on staging with and without onboarding. LGTM! 🚢

PR #169, opened 2025-08-26
Replace base URL with new threadID when available

### LINKED ISSUE #162: Append new conversation url with thread id when available
As reported in testing:
> when I'm in a new conversation thread chatting and then select the "new conversation" button nothing happens. When I click to another conversation thread and then click the "new conversation" button it works.

This occurs because when a user is in a "new conversation" the URL is the base app url, and the "New conversation" button is simply a link to this same URL.

Once a conversation has a thread id, the url should be updated with this id.

This will allow users to copy the link to the thread if desired, without needing to navigate away from the page and reload the thread. More importantly, this will allow users to start a new conversation.

LanesGood:
Adds threadID to URL as soon as new conversation gets an ID.

The main impetus for this change is to enable the "New conversation" button for users who are currently chatting in a new conversation (see #162)

Closes #162

### DIFF
--- a/app/app/page.tsx
+++ b/app/app/page.tsx
@@ -4,11 +4,21 @@ import { useEffect } from "react";
 import useChatStore from "@/app/store/chatStore";
 
 export default function Home() {
-  const { reset: resetChatStore } = useChatStore();
+  const { reset: resetChatStore, currentThreadId } = useChatStore();
 
   useEffect(() => {
     resetChatStore();
   }, [resetChatStore]);
 
+  useEffect(() => { // Update URL as soon as threadID is available
+    if (currentThreadId) {
+      const newUrl = `/app/threads/${currentThreadId}`;
+      window.history.replaceState(
+        { ...window.history.state, as: newUrl, url: newUrl },
+        "",
+        newUrl
+      );
+    }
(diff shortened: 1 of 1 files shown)


LanesGood:
@danielfdsilva on second look, it does look like this change is also needed because in your PR, while the URL is updated the app is still (I think) using the component at `app/(chat)/page.tsx`, but doesn't force a change to load the component at `app/(chat)/threads/[id]/page.tsx`. Without this change, what I'm seeing on `develop` with #151 merged in is that while the URL is updated and the new chat is added to the sidebar, I'm actually no longer in my chat and don't see any of the returned messages unless I reload/click in and out of that new thread:

_Screen recording from `develop`_

https://github.com/user-attachments/assets/ca18cbd2-d54e-4c17-b910-77c3f6945bfa

danielfdsilva:
@LanesGood I was able to replicate the error, but it's not related to the url.
The problem is that the langchain message that's returned no longer has a timestamp (cc @batpad @leothomas - Any idea what happened?)  and so there's an error, preventing the message from showing up.
Fix is in this PR: https://github.com/wri/project-zeno-next/pull/181

PR #174, opened 2025-08-27
Filter layer widgets from Go to insights

01painadam:
The new data layer cards are message.type === widget so they show up in the Go to insight drop-down.

image

This PR filters these out and updates the default title if one is not present


01painadam:
Exactly @LanesGood, it should remove them from the `Go to Insight` dropdown

PR #175, opened 2025-08-27
Update/theme landing

LanesGood:
Updates theme values with the colors from the recently shared design system files, and updated those where needed.

Also adds a small landing page tweak for the visibility of the hero page text and alignment with design.


LanesGood:
Some style changes here that have no effect outside of the landing page; merging to avoid excess review time. cc @01painadam @kamicut

PR #176, opened 2025-08-27
Set dataset card as active by default

01painadam:
When a dataset card is returned in chat, the map layer is set active by default. 

Screenshot 2025-08-27 at 5 23 09 PM

cc: @faustoperez since we discussed this


01painadam:
@LanesGood Yeah true - we will need to allow two layers active (only ever the primary layer + its specific context layer)... which will require some specific busines logic/exception handling. 

Shall I revert that commit for now?

PR #181, opened 2025-08-28
Handle missing timestamp in langchain message

danielfdsilva:
There as some refactoring to simplify the logic!


LanesGood:
I can confirm that this fixes the main issue mentioned in #182

PR #184, opened 2025-08-28
Anonymou UI

danielfdsilva:
Implement anonymous UI messages and chat input disable.
Classic map page in progress at `/classic`

### DIFF
--- a/app/(chat)/layout.tsx
+++ b/app/(chat)/layout.tsx
@@ -11,7 +11,7 @@ import { Sidebar } from "@/app/sidebar";
 import useMapStore from "@/app/store/mapStore";
 import useContextStore from "@/app/store/contextStore";
 import PageHeader from "@/app/components/PageHeader";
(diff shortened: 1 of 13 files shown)


danielfdsilva:
@kamicut I'd say this can be reviewed and merged. We can continue the needed work on another branch.

- The map is at `/app/classic` and features the same layer selector from the context panel.
- The legend components are created but are mainly static. Waiting for @faustoperez on final designs and we still need to understand where the data for the legend comes from. Is it hardcoded like the layer options? cc @yellowcap 
- The functionality for anonymous users is still in, but unless you set the user as anonymous with the appropriate method, it will never trigger. I did not remove the initial "login/dismiss" prompt, as I don't know what it should be replaced with (And I think there's someone working that part)
- I had to wrap the main chat and threads pages in a nextjs group so their layout wouldn't affect the classic page.

kamicut (approved):

PR #193, opened 2025-09-02
Prevent file dialog from opening twice

danielfdsilva:
The button was triggering the dialog and then the event bubbled up to the parent element which cause the dialog to open again.


LanesGood (approved):
Works as expected. Thanks! Hadn't realized that the visuallyHidden component also worked to prevent bubbling - nice fix.

PR #41, opened 2025-07-23
Add geometry draw

vgeorge:
Contributes to #24.

This is WIP.

Changes:

- Add TerraDraw library
- Added draw controls to the map
- Refactored mapStore to use [Zustant slices](https://zustand.docs.pmnd.rs/guides/typescript#slices-pattern)
- Added drawAreaSlice

The next step is to add custom areas layer to render drawn areas on the map.

cc @kamicut @danielfdsilva @wrynearson @LanesGood

PR #214, opened 2025-09-08
Clear mapstore when loading new thread

### LINKED ISSUE #196: Clear map context when switching threads
When switching between threads or starting a new conversation, the mapstore should be cleared to avoid confusion and overlap between previous area contexts. We should also clear the active datasets. Otherwise, the map is very messy:

Image

LanesGood:
This PR adequately closes #196 - map layers and AOIs are removed from the map when a new thread is created or a previous one selected.

One remaining issue: the `flyToCenter` function doesn't seem to be working with this change. @kamicut or @danielfdsilva would be great if you could take a look.

https://github.com/user-attachments/assets/a4f74cfc-9493-40c3-aeba-d625652f7217


kamicut (approved):

LanesGood:
Confirming that this works, but want to ping @vgeorge for a quick look at this as it appears that he was the one to set up the mapStore reset function when working on https://github.com/wri/project-zeno-next/pull/41. @vgeorge any concerns that not resetting the `mapRef` may affect AOI drawing/selection?

vgeorge:
@LanesGood I think that was a bug. If the map is mounted and we just want to clear its state, we can do so without touching the mapRef.

PR #215, opened 2025-09-08
Add user dropdown

LanesGood:
This PR:
- Adds the user dropdown UI
- Uses the `clearAuth` method from authStore as the logout function is not yet complete.
- Links to the ~home~ "settings" page ~and non-existent "about" page~

image


faustoperez:
> @faustoperez would you be able to provide a design review? This wasn't mocked up on figma, just here

Looks good, but I'm not sure the 'Homepage' and 'About' links belong there. 🤔 . I suggest:

- Linking the logo to the homepage
- The options for the user dropdown would be:
    - Settings
    - Log out (maybe in red as the delete conversation option)
    
    
image

Then we can add 'About' and 'Help' to the user dashboard sidebar.

Here are the user dashboard designs:
https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/WRI_LCL_ProjectZeno?node-id=2676-16370&t=IthQoK96JIDpLYHG-1

LanesGood:
Thanks @faustoperez - I've removed the homepage and about links, and linked the header logo + page title as you suggested. Logout is now in "danger" color.

@kamicut I replicated the login overlay click handler for the Login button in the page header; let me know if you think we should place these now exported constants in a more central place.

PR #234, opened 2025-09-11
Add GA events

01painadam:
Adds further GA4 event tracking and update cookie consent banner copy.


LanesGood (approved):
Looks good - I assume you're testing these events in GA.

PR #195, opened 2025-09-02
Add uploaded feature to map and context

danielfdsilva:

PR #257, opened 2025-09-13
Add drawn custom area to map context immediately after drawing

### LINKED ISSUE #140: Add custom area to context and map immediately
After uploading or drawing an AOI on the map, that area should be immediately added to context and persist on the map. Currently these are removed from the map and not added to context, and a user has to go to the "custom area" selection drop-down.

cc @kamicut @vgeorge

LanesGood:
Closes #140 - adds drawn areas to the map and context after confirming a successfully drawn area. Follows the pattern of #195.

NB - The fixes for drawing in #255 are needed to test this PR.


app/components/MapAreaControls.tsx
```diff
@@ -90,6 +94,46 @@ function MapAreaControls() {
     };
   }, [clearSelectionMode, isDrawingMode, cancelDrawing]);
 
+  const handleConfirmDrawing = async () => {
+    try {
+      const result = await confirmDrawing();
+      if (!result) return;
+      console.log(result);
```
kamicut, line 101:
```suggestion
```

kamicut (approved):

PR #269, opened 2025-09-15
Fix bad setstate, update welcome prompts

LanesGood:
This PR:
- Removes the functionality of clicking a scrolling prompt to set that prompt in the input box (fixes bad setstate)
- Removes the hardcoded "sample prompts" from the homepage, in favor of the prompts from the fetchPrompts() function


LanesGood:
This implements the changes mentioned in https://github.com/wri/project-zeno-next/pull/168#issuecomment-3287366908 just for the welcome prompts and scrolling prompt marquee cc @01painadam @wrynearson (given that the other pr is not as top of a priority and still needs changes)

PR #273, opened 2025-09-15
Send only new context to the llm

danielfdsilva:
Context is only sent to the llm once, with the user message right after the context is selected.

### DIFF
--- a/app/components/widgets/DatasetCardWidget.tsx
+++ b/app/components/widgets/DatasetCardWidget.tsx
@@ -1,5 +1,4 @@
 import { DatasetInfo } from "@/app/types/chat";
-import { useEffect } from "react";
 import useContextStore from "@/app/store/contextStore";
 import { DatasetCard } from "@/app/components/DatasetCard";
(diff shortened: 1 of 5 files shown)


app/components/widgets/DatasetCardWidget.tsx
```diff
@@ -18,20 +17,6 @@ export default function DatasetCardWidget({ dataset }: DatasetCardWidgetProps) {
...
-        contextType: "layer",
-        content: dataset.dataset_name,
-        datasetId: dataset.dataset_id,
-        tileUrl: dataset.tile_url,
-        layerName: dataset.dataset_name,
-      });
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []); // Run only once on mount
-
```
danielfdsilva, line 34: The dataset context was being set whenever the dataset card was mounted, which is a very different behaviour from the other context pieces, so I move it to the appropriate context action.
This was also needed for this feature to work

kamicut (approved):

PR #280, opened 2025-09-15
make label hit box and clear fill

kamicut:
- Make name label the hit box
- Remove fill area and color polygon with a stroke (#271)

https://github.com/user-attachments/assets/0af3e416-bb63-4c54-9f7e-013289813bb7

### DIFF
--- a/app/components/map/layers/HighlightedFeaturesLayer.tsx
+++ b/app/components/map/layers/HighlightedFeaturesLayer.tsx
@@ -1,11 +1,5 @@
 import React, { useEffect, useState, useCallback, FC } from "react";
-import {
-  Source,
-  Layer,
-  Marker,
-  useMap,
-  MapMouseEvent,
-} from "react-map-gl/maplibre";
+import { Source, Layer, Marker } from "react-map-gl/maplibre";
 import { Tag } from "@chakra-ui/react";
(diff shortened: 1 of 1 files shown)


LanesGood (changes requested):
Let's update the fill color to use the var(--chakra-colors-primary-solid) for selected and var(--chakra-colors-neutral-emphasized) for unselected

LanesGood (approved):
I made a few follow changes here:
1. Changed the colors for the inContext/unselected areas to match the theme colors (though am using the direct hex value, rather than fussing with calling theme values via Chakra hooks)
2. Updated the label to be `solid` when selected to stand out, and `subtle`/`surface` when unselected/hovered.
3. Removed the dashed area polygon line as this is not present in the designs; area polygons always have a solid line, while the bbox rectangle around a polygon is dashed when unselected and unhovered.
4. Reordered the layers on the Map component to ensure that highlighted feature layers are rendered above the dynamic tile layers 
5. Also updated the legend opacity control to ensure it's not cut off by the legend overflow hidden rule.  [off topic change, but related to map changes]

Approving this and will update the main ticket as well

https://github.com/user-attachments/assets/9ff9203e-cb39-4140-9fad-5a2e4191569d

PR #282, opened 2025-09-15
Feature/rating messages

kamicut:
Implements #21 - although I will need help with this one on the server side @batpad / @sunu 
I'm not sure why `trace_id` sometimes comes attached to the message and sometimes as separate messages, so I had to implement handlers to keep trace_id as a "pending message" to attach it to previous messages. To test this logic I will have to have access to Langfuse or to the output of the feedback.

@LanesGood could you please review the "thumbs down - leave a review" input box I made?

### DIFF
--- a/app/components/DebugToastsPanel.tsx
+++ b/app/components/DebugToastsPanel.tsx
@@ -54,7 +54,7 @@ function DebugToastsPanel({ enabled }: { enabled?: boolean }) {
             toaster.create({
               title: "Warning",
               description: "This is a warning toast",
-              type: "Warning",
+              type: "warning",
               closable: true,
               duration: 3000,
             })


LanesGood:
@kamicut I'd suggest that we follow the proposed feedback popout instead of an in-chat input box. I've mocked this up here and provided a screen recording and code.

https://github.com/user-attachments/assets/305e028b-3d71-4651-911e-d0d6bd38605e


Full code for this is pasted below. Requires the import of `Portal` and `Popover` from chakra, and note a little strange <box> thrown in there (apparently necessary for a tooltip to wrap a popover trigger given they both use popper.js). This would replace the current Tooltip/Icon for "Bad response" and the Box containing the TextArea form (line 255 - 299).


<details>
<summary>Expand to see code</summary>

```tsx
            <Popover.Root
                open={feedbackOpen}
                onOpenChange={(e) => setFeedbackOpen(e.open)}
                positioning={{ placement: "bottom-end" }}
              >
                <Tooltip content="Bad response">
                  <Box display="inline-block">
                    <Popover.Trigger asChild>
                      <IconButton
                        variant="ghost"
                        size="xs"
                        disabled={isRating || !message.traceId}
                        onClick={async () => {
                          await rateMessage(-1);
                          setFeedbackOpen(true);
                        }}
                      >
                        <ThumbsDownIcon />
                      </IconButton>
                    </Popover.Trigger>
                  </Box>
                </Tooltip>
                <Portal>
                  <Popover.Positioner>
                    <Popover.Content>
                      <Popover.Body>
                        <Textarea
                          value={feedbackText}
                          onChange={(e) => setFeedbackText(e.target.value)}
                          placeholder="Tell us what went wrong (optional)"
                          size="sm"
                          rows={3}
                        />
                        <HStack justify="flex-end" mt="2" gap="2">
                          <Button
                            variant="ghost"
                            size="xs"
                            onClick={() => {
                              setFeedbackOpen(false);
                              setFeedbackText("");
                            }}
                          >
                            Cancel
                          </Button>
                          <Button
                            size="xs"
                            disabled={isRating}
                            colorPalette="primary"
                            onClick={submitFeedback}
                          >
                            Send feedback
                          </Button>
                        </HStack>
                      </Popover.Body>
                      <Popover.CloseTrigger />
                    </Popover.Content>
                  </Popover.Positioner>
                </Portal>
              </Popover.Root>
```

</details>

LanesGood (approved):
Works for new/active conversations!

We should note in the main ticket that the rating does not yet work for past threads.

PR #288, opened 2025-09-16
navigate to rw logout

kamicut:
Navigates the page explicitly to RW logout page. Ideally the callbackUrl throws us back to the application, but a fallback would be to have a link back to the homepage here @01painadam

### DIFF
--- a/app/components/PageHeader.tsx
+++ b/app/components/PageHeader.tsx
@@ -143,13 +143,16 @@ function PageHeader() {
           </Menu.Root>
         ) : (
           <Button
+            asChild
             variant="solid"
             colorPalette="primary"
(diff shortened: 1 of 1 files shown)


kamicut:
@tanderegg in this PR we delete our cookies and send a GET request to api.rw.org/auth/logout but we still don't get logged out of the proxy. Clicking on login will skip the proxy and act as if we've logged in on RW.

kamicut:
turning this back into a draft until the RW gateway implements some form of callback

tanderegg:
@kamicut FYI, the callback on logout is now implemented, you can redirect via e.g. `https://api.resourcewatch.org/auth/logout?callbackUrl=https://www.globalnaturewatch.org/`.

PR #291, opened 2025-09-16
Improve Pie charts

01painadam:
Readability enhancements to pie charts:
- Increase legend label and text size
- Orient segment to 12oclock
- Use dataset specific palette
- ordering the labels to match the map legend order (and a more visibly pleasing order

Screenshot 2025-09-16 at 8 14 19 PM

Screenshot 2025-09-16 at 8 14 57 PM

NOTE: the palette should probably be constructed from the DATASET_CARDS fixture as a single source of _non-agent_ truth.


01painadam:
We should probably get the agent to agg all non-natural landcover for Natural Lands. It ends up a bit of a mess (esp. since they all user the grey palette anyway)

01painadam:
Labels now ordered

Screenshot 2025-09-16 at 8 41 44 PM
Screenshot 2025-09-16 at 8 41 57 PM

LanesGood (approved):
Oooooh really nice! 🧑‍🎨 🎨 

I don't think this is related, but strangely I don't see a legend for the "TCL by Driver" layer:
Image

Also, I don't know why the main item here wouldn't be rendered in the chart legend. Again, I don't think related to your code:
Image

For later enhancement, I'm guessing this formatting can also be applied to bar/scatter/area charts as needed.

01painadam:
Oh, good catch -  didnt test that as I thought TCL Drivers wasnt merged. Let me take a quick look

01painadam:
@LanesGood I see the legend 🤷 

I did fix the default colour being injected (it was a typo on one of the keys) and added % to the pie

Issue #290, opened 2025-09-16
Form doesn't validate on /onboarding


kamicut:
We've seen instances where if the form is filled out with a keyboard (without an explicit dropdown select) the form validation does not make the Continue button valid.

LanesGood:
Closed by #292

PR #292, opened 2025-09-16
onboarding changes

### LINKED ISSUE #259: block organic access to /onboarded after registration
### Description
Users can still navigate to `/onboarding` directly after finishing sign-up, which is confusing and can cause state inconsistencies.

After completing registration, users should not be able to reach the onboarding flow. Unauthenticated users should be redirected to sign-in.

### Acceptance Criteria
- Navigating to /onboarding once already registered in redirects to /app
- Navigating to /onboarding while not already registered continues to onboarding normally.

kamicut:
- change from `onClick` to onValueChange #290 
- make topics not required
- fix #259

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -36,11 +36,12 @@ export async function middleware(request: NextRequest) {
     return NextResponse.next();
   }
 
-  // General guard for /app and /onboarding
+  // General guard for /app, /onboarding, and /dashboard
   const isOnboarding = pathname.startsWith("/onboarding");
   const isApp = pathname.startsWith("/app");
+  const isDashboard = pathname.startsWith("/dashboard");
 
-  if (isOnboarding || isApp) {
+  if (isOnboarding || isApp || isDashboard) {
     const authCookie = getAuthTokenFromRequest(request);
 
     // If not authenticated, redirect to WRI login


middleware.ts
```diff
@@ -31,14 +31,16 @@ export async function middleware(request: NextRequest) {
...
     return NextResponse.next();
   }
 
-  // Guard restricted routes under /app
-  if (pathname.startsWith("/app")) {
+  // General guard for /app and /onboarding
+  const isOnboarding = pathname.startsWith("/onboarding");
+  const isApp = pathname.startsWith("/app");
+
+  if (isOnboarding || isApp) {
```
LanesGood, line 43:
```suggestion
  const isApp = pathname.startsWith("/app");
  const isDashboard = pathname.startsWith("/dashboard");

  if (isOnboarding || isApp || isDashboard) {
```

Would this also solve the need to protect the Dashboard page route?

LanesGood (approved):
Looks correct, and I can confirm that onboarding is protected. @kamicut one question above about protecting dashboard route - otherwise, good to go.

middleware.ts
kamicut: yes thank you!

PR #313, opened 2025-09-17
Fix mobile layout flash

LanesGood:
CSS Hotfix for flash of extra space in layout coming from mobile layout render.

### Current challenge:

https://github.com/user-attachments/assets/cbe0dae5-b393-4315-a951-3009947c5e04

### Why this is happening:
Both the mobile and desktop layouts are rendered by default on the server with SSR. The Chakra `useBreakpointValue` hook can't fire until the page is initially rendered on the client, and so both components are loading and taking up space. This PR attempts to set a 0 height on both the layout wrappers, and then set the correct heights on load.

### DIFF
--- a/app/app/(chat)/layout.tsx
+++ b/app/app/(chat)/layout.tsx
@@ -6,6 +6,7 @@ import {
   Drawer,
   Portal,
   IconButton,
+  useBreakpoint,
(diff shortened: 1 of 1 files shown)


LanesGood:
This can be merged

PR #328, opened 2025-09-24
Update ga triggers

01painadam:
- Adds a manual area select trigger
- Update response feedback triggers to be distinct


01painadam:
@kamicut this can be released to prod too once reviewed

kamicut (approved):

PR #334, opened 2025-09-25
Share Threads

### LINKED ISSUE #78: Thread Sharing
Allow users to share read-only threads by making them public.

- Design affordance to "Make thread public"
- Users can then copy & share link
- Any user can the read-only thread (whether logged in or not)

Note: what happens when a thread is deleted? What does a user navigating to that link see?

LanesGood:
Allow users to share threads.

Creates a "Share Thread" dialog accessible from the thread actions menu found in both the sidebar thread options and the Chat Panel Header.

https://github.com/user-attachments/assets/353da5e6-a610-4488-b270-23407667f280

@01painadam the "This is legal copy" text in the dialog needs a review. We may also want to add text noting that users do need to be registered and signed in to see threads, at least for the closed/limited beta.

Closes #78

### DIFF
--- a/app/components/ThreadShareDialog.tsx
+++ b/app/components/ThreadShareDialog.tsx
@@ -115,8 +115,7 @@ function ThreadShareDialog(props: ThreadShareDialogProps) {
                 Sharing creates a public, view-only link to this conversation.
                 Make sure it contains no personal or sensitive information.
                 You can switch Visibility back to private later.
-                <br />
-                Learn more in our{' '}
+                For more information, read our{' '}
                 <Link
                   textDecoration="underline"
                   textDecorationStyle="dotted"
--- a/app/components/ThreadShareDialog.tsx
+++ b/app/components/ThreadShareDialog.tsx
@@ -20,6 +20,7 @@ import {
   CopyIcon,
   CheckIcon,
 } from "@phosphor-icons/react";
+import { sendGAEvent } from "@next/third-parties/google";
 import ChatDisclaimer from "./ChatDisclaimer";
 
 interface ThreadShareDialogProps {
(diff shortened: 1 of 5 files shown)


app/components/ThreadShareDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type ValueChangeDetails = { value: string[] };
+
+function ThreadShareDialog(props: ThreadShareDialogProps) {
+  const { threadId, isPublic, onShare, isOpen, onOpenChange } = props;
+  const shareUrl =
+    typeof window !== "undefined"
+      ? `${window.location.origin}/app/threads/${threadId}`
+      : threadId;
```
LanesGood, line 40: @kamicut I am not sure how to access the Vercel env vars here to correctly set the share URL prefix

kamicut (approved):

LanesGood:
Thanks for updating the copy and catching the menu bug @01painadam  - I've pushed a change to fix this.

I like the idea of a signifier on shared chats.

I see that in ChatGPT, when a thread is shared there's no immediately visible indicator of sharing. When a user clicks "share" again, they can see this modal:
image

I'm curious if we are implementing the same behavior - where the public share may be only up to a certain traceID/checkpoint, but would need to confirm. If this is the case, a "thread is public" indicator is a bit misleading, as the public part of the thread is only up to that share point.

01painadam:
> where the public share may be only up to a certain traceID/checkpoint, but would need to confirm. If this is the case, a "thread is public" indicator is a bit misleading, as the public part of the thread is only up to that share point.
@batpad is this correct?

PR #335, opened 2025-09-29
GA4 fixes

kamicut:
- Move analytics to single component
- Fix signature for login event
- Remove redundant logging in event


01painadam (approved):
LGTM - I did see the old banner the first time I loaded but after clearing my storage I don't see it now

PR #342, opened 2025-10-06
Initial prompt disclaimer box

01painadam:
Now we have more certainty over the rollout, re-implementing @LanesGood's disclaimer box on the initial message. 

Screenshot 2025-10-29 at 5 38 23 PM

### DIFF
--- a/app/components/ChatMessages.tsx
+++ b/app/components/ChatMessages.tsx
@@ -1,16 +1,18 @@
 "use client";
-import { Fragment, useEffect, useRef } from "react";
-import { Box } from "@chakra-ui/react";
+import { Fragment, useState, useEffect, useRef } from "react";
+import { Box, Text, Link } from "@chakra-ui/react";
 import useChatStore from "@/app/store/chatStore";
 import MessageBubble from "./MessageBubble";
 import Reasoning from "./Reasoning";
 import SamplePrompts from "./SamplePrompts";
+import ChatDisclaimer from "./ChatDisclaimer";
 
 const LANDING_PAGE_VERSION = process.env.NEXT_PUBLIC_LANDING_PAGE_VERSION;
 
 function ChatMessages() {
   const containerRef = useRef<HTMLDivElement>(null);
   const { messages, isLoading } = useChatStore();
+  const [displayDisclaimer, setDisplayDisclaimer] = useState(true);
 
   // Auto-scroll to bottom when new messages are added or loading state changes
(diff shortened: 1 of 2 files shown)


app/components/ChatMessages.tsx
```diff
@@ -53,6 +55,40 @@ function ChatMessages() {
         const isFirst = index === 0;
         return (
           <Fragment key={message.id}>
+                        {displayDisclaimer && (
```
LanesGood:
```suggestion
                        {isFirst && displayDisclaimer && (
```
Only display disclaimer on first message - otherwise will show on every message like so:

Image

In the initial implementation proposed, this also was only on the non-public version: https://github.com/wri/project-zeno-next/pull/258/commits/992210063c788c1e483de1410def49234e3b7df8

However, I think we could just proceed with this included on any version, and change the text or add/remove as needed

app/components/ChatMessages.tsx
```diff
@@ -1,16 +1,18 @@
...
 import Reasoning from "./Reasoning";
 import SamplePrompts from "./SamplePrompts";
+import ChatDisclaimer from "./ChatDisclaimer";
 
 const LANDING_PAGE_VERSION = process.env.NEXT_PUBLIC_LANDING_PAGE_VERSION;
 
 function ChatMessages() {
   const containerRef = useRef<HTMLDivElement>(null);
   const { messages, isLoading } = useChatStore();
+  const [displayDisclaimer, setDisplayDisclaimer] = useState(true);
```
kamicut, line 15: Should setDisplayDisclaimer write to a cookie or local storage?

01painadam: Good catch - eventually we will want to set this in local so it doesnt pop up each time... but for now I think its ok if it pops up each session as a reminder.

app/components/ChatMessages.tsx
```diff
@@ -53,6 +55,47 @@ function ChatMessages() {
...
+                <br />
+                  Your feedback is critical to improving Global Nature Watch. Complete{" "}
+                  <Link
+                    color="primary.solid"
+                    textDecor="underline"
+                    href="https://surveys.hotjar.com/860def81-d4f2-4f8c-abee-339ebc3129f3"
+                  >
+                    this survey
+                  </Link>{" "}
+                   or us at{" "}
```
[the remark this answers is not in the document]
01painadam: good bot

PR #344, opened 2025-10-13
use osano for cookie consent

kamicut:
This changes the consent for GA4 to use Osano. Since the script is in [discovery mode](https://docs.osano.com/hc/en-us/articles/22472145619988-Compliance-Modes-Listener-Permissive-Strict) it doesn't trigger a consent banner. Is this something we can try to change in the dashboard @01painadam to see how the banner behaves?


01painadam:
Great - I've requested the test deploy url be added to Osano so we can test

01painadam (approved):
LGTM, I'm seeing the banner on /app and the landing page.

Just to confirm, we should still see GA events being triggered if users accept?

PR #365, opened 2025-11-04
Add CopySelectionTooltip component

### LINKED ISSUE #359: Copy text disclaimer
### Description

Implement a custom tooltip that is surfaced when a user highlights text to copy. This makes clear to the user they are extracting generative information, which requires their own verification.

Image

See design: https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/Global-Nature-Watch?node-id=4277-10276&t=ATFtZY9mdFck27R4-4

kamicut:
PR to close #359 

image

@LanesGood @faustoperez the designs show a yellow AI sparkle, is that `secondary.500`?

### DIFF
--- a/app/components/CopySelectionTooltip.tsx
+++ b/app/components/CopySelectionTooltip.tsx
@@ -1,0 +1,141 @@
+"use client";
+import { Box, Button, Flex, Icon, Text } from "@chakra-ui/react";
+import { CopyIcon, SparkleIcon } from "@phosphor-icons/react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
+
+interface CopySelectionTooltipProps {
+  children: React.ReactNode;
+  /** Optional: limit tooltip to render only when true */
+  enabled?: boolean;
+  /** Message shown inside the tooltip */
+  notice?: string;
+}
+
+/**
+ * Wrap content with a listener that surfaces a tooltip when the user selects text.
+ * The tooltip includes a notice about AI-generated content and a Copy button
+ * that copies the current selection to the clipboard.
+ */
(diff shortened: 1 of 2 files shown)


01painadam:
Noice! LGTM with a few small additions:

- On ctrl/cmd+c also dismiss the component
- I think the component should be positioned in the centre of the pane and not expand/shrink as the text selected (since thsi causes an overflow outside the pane)
- I think the sparkle needs a new colour in theme: `lime.500.value = "#C3D16F"` and should be solid fill

app/components/CopySelectionTooltip.tsx
```diff
@@ -0,0 +1,141 @@
...
+          borderRadius="md"
+          shadow="md"
+          borderWidth="1px"
+          borderColor="border"
+          maxW={{ base: "calc(100% - 16px)", md: "560px" }}
+          p="2.5"
+          ref={tooltipRef}
+        >
+          <Flex align="center" gap="2">
+            <Icon as={SparkleIcon} color="fg.muted" />
```
LanesGood:
```suggestion
            <Icon as={SparkleIcon} color="secondary.500" />
```
@kamicut yes, secondary.500

01painadam (approved):
LGTM, made a few suggestions above but not blocking!

kamicut:
I've added:
- cmd+c to dismiss
- center tooltip
- I've enabled "copy text" on mobile, however I'm not too happy with this because it negatively interacts with the draggable drawer

01painadam:
Nice, let's merge! @kamicut could you add a follow up ticket in the project to solve the mobile issue - I don't think we need to solve now but good to keep track

PR #366, opened 2025-11-04
Add text-select copy disclaimer

### LINKED ISSUE #359: Copy text disclaimer

01painadam:
### Summary
This PR adds a tooltip that appears when users highlight text in assistant responses. The tooltip displays a disclaimer reminding users to verify AI-generated content before using it in their work, along with a convenient copy button.

chrome-capture-2025-11-4

Behaviour:
- When you select/highlight text in an assistant's message, a tooltip appears below the selection
- The tooltip is centered below on the selected text and stays within the chat pane boundaries
- Only one tooltip appears at a time (selecting text in a different message closes any existing tooltip)
- The tooltip closes when users copy the text with ctrl/cmd+c, click away, or select different text

Adds:
 - Custom hook `useSelectionTooltip.ts` that manages tooltip state and positioning logic
 - Tooltip component `SelectionTooltip.tsx`

Updates:
 - `ChatMessages.tsx` uses the custom hook and renders a single tooltip instance
 - `MessageBubble.tsx` sends `onMouseUp` events to parent for tooltip triggering

### Open questions
- Is there a better way to position the tooltip in the pane?
- is its prominent enough? I feel like it blends in with the chat pane
- should ctrl+c dismiss or set button text to `Copied`? (and could clicking `Copy` dismiss?)
- should there be a button to dismiss the component?

### DIFF
--- a/app/components/ChatMessages.tsx
+++ b/app/components/ChatMessages.tsx
@@ -59,7 +59,6 @@ function ChatMessages() {
         onClose={onClose}
         target={tooltipTarget}
         onCopy={handleCopy}
-        containerRef={containerRef}
       />
       {messages.map((message, index) => {
         // Check if this message is consecutive to the previous one of the same type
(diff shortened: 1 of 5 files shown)


01painadam:
Closing since its a dupe :)

PR #393, opened 2025-12-19
Update cookie banner styles

LanesGood:
Updates cookie consent banner styles to be more muted:

Desktop
image

Mobile: 
image

Also deletes unused cookie consent component and store to avoid confusion.


01painadam (approved):
Much better!

Issue #363, opened 2025-11-03
Implement MVP persistent reasoning component


01painadam:
## Summary
The goal is to expose a clear, human-readable thread from the user’s prompt to the final result, while the header cycles through live steps and the expanded view forms a permanent, audit-ready timeline.

[Design](https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/Global-Nature-Watch?node-id=4273-7654&t=ATFtZY9mdFck27R4-4)

## Scope
IN
- reasoning header text cycles through steps
- add shimmer to header text
- update 'processing' animation to match design
- add "timeline"
- persist reasoning component with "X steps completed"

BONUS
- Add "reasoned for X seconds on completion

OUT OF SCOPE
- Enhanced reasoning inside component - use existing assistant messages

01painadam:
@faustoperez to what extent do you feel this is made redundant by the _"how was this generated"_ component?

faustoperez:
> @faustoperez to what extent do you feel this is made redundant by the "how was this generated" component?

I think they are separate things: reasoning is for the model to interpret the prompt and its multiple things, not just the chart. The inspection panel focuses on the actual code and steps used to generate the chart so that people can replicate it elsewhere. That was my understanding. @yellowcap Let me know if this makes sense or if I'm missing anything.

yellowcap:
Agreed, we should still have a nicer way to show what tool calls the agent works through and persist that at the end in a small collapsed component.

Issue #362, opened 2025-11-03
Explore reasoning enhancements


01painadam:
## Summary

Explore how we can update the expandable _reasoning_ component to expose more information about what the agent is doing under the hood in order to improve transparency, and ensure a clear thread-of-continuity from prompt to final result.

It is unclear how the agent's internal steps can be surfaced as a clear, human-readable timeline that persists after completion (e.g. “Reasoned for X seconds, N steps completed”). This spike will determine what information we can safely expose, how to structure and render it for users, and the FE-BE contract needed to stream and display it reliably.

This enables: trustworthy, auditable results; easier debugging/support; and a foundation for future features (e.g. shareable run transcripts, export, or QA tooling).

[Designs](https://www.figma.com/design/N20cpSWt9kWZWILQOcsdIT/Global-Nature-Watch?node-id=4273-7654&t=ATFtZY9mdFck27R4-4)

## Questions to answer
1. What can we expose?
Which step-level fields (status, timestamps, “what/result/why”, sources, code snippets, errors) are safe, useful, and non-sensitive? What redaction and truncation rules are required?

2. How should it be formatted?
What schema supports a human-readable timeline, live step cycling, and a persistent summary? How do we represent statuses, ordering, nested sub-steps, and optional sections (Sources, Code) across streaming and final payloads?

## Outputs
ADR with:
- List of fields we will expose per step, with examples and rationale.
FE renderer notes
- Mapping of schema → UI (collapsed header, timeline entries, badges, copy/expand for code, links...

01painadam:
@yellowcap FE is now able to surface this

PR #396, opened 2026-01-14
Persistent Reasoning Improvements

kamicut:
https://github.com/user-attachments/assets/91afd065-434f-4353-9b7b-0a372b31b568

Here are a few improvements to the reasoning component
- Persistence: every user query has its associated reasoning component
- A dropdown shows the message data from the API. @yellowcap not sure how we style this or what is most useful to show here.
- Shimmer text and duration @LanesGood I did my best here but it could use some improvements
- **Todo**: fix css bug where clicking on the dropdown brings you to the end of the chat


yellowcap:
Closes #362 and #363

yellowcap:
Looks great to me, the "raw" json rendering is useful for me, that is a bit like a debug log. But not sure if this is a good thing from a design perspective @LanesGood .

LanesGood:
This is looking really good! 

The original design linked in the enhancements ticket has some but not all of the reasoning text formatted as "raw" json:
image

I'd be happy to help with aligning more closely to the designs which I think help contain the component a bit more and keep it distinct from the chat. @kamicut let me know if you are working on these updates or if I should take a shot.

LanesGood:
I've updated the styles for the reasoning component and think this is quite close to the designs. Screenshot of my latest commit: 
image

However we should further establish how the content is presented.

The original design from @faustoperez proposes that the majority of the text for each step be returned as regular text, and I don't see a need for this to be inserted in a pre-formatted code block.

I see that we are currently rendering everything returned by the tools output; the `codeact_parts` object is always in base64 and I'm not sure if it should be rendered, as its a lot of information and is present in the provenance sidebar.

Each tool has its own regular outputs; I think we can strip this down to what is needed from those outputs and render it as mainly text. Some of the data for the "Generate Insights" tool _could_ go into a code block for the charts data, but I don't know if this is really necessary. As @01painadam points out in #362, there needs to be a good understanding of what we should expose from each step. I'm going to change the status on that ticket to "developing" as I don't think this PR addresses this yet.

As a potential solution, I'd suggest that we just render the `tool.content` part of each response. I've attempted to do that locally, and then process the markdown via the react-markdown library, but something is strange with the output (capturing markdown as code)
image

I'm pushing up my commented code to see if you can help fix the rendering @kamicut 

---
Separately, I also note a **bug** - when I leave a conversation and then come back to it later, the Reasoning component is squashed into a single instance. For example, if I submit two prompts on a conversation, I do correctly get two separate "Reasoned for x seconds" with the unique steps in the accordion for each. However, if I leave and return to that chat, I just have one "Reasoned for x seconds" with all steps listed in aggregate.

---
Finally, noting that in the original designs, we do present the idea that a user can expand the "Reasoning" steps while the tool steps are running; in this branch, a user must wait until all steps are complete.

01painadam (changes requested):
Nice, I like the transparency re: what's going on in the background. A couple of thoughts...

- Should we make the reasoning component move down the chat pane with the messages (so its always at the bottom). I say this because I doubt users will scroll back up to look at the reasoning component. Any opinions @faustoperez @LanesGood 
- The above should fix the issue where the chat pane jumps to the bottom when you open it. Fix this.
- I think we need to come up with a way to better communicate the contents inside the expanded component (the json info isn't gonna help many users). For now, I'd stick to a simple checklist of things done (and, even better if possible) a single statement of the outcome eg. 

```
Reasoned for Xs:
    Selecting Area: France selected ✅
    Selecting Dataset: Tree cover loss 2015-2020 ✅ 
    Pulling data: successful ✅ 
    Generating Insight: done ✅ 
```

LanesGood:
- I do think we need to fix the scroll-to-bottom issue, and also think that we could explore positioning the reasoning component to be below the message; however, we'll want to ensure that this doesn't obstruct or distract from the chat's proposed next steps
- I like the proposed checklist of items. We could also still provide more detail in a nested accordion item if needed. @kamicut do you think it's straightforward to convert the step into a summarized, single line? I'm not sure if this would require a dictionary, or if there's an element that can be picked that is consistently summarizing the step

yellowcap:
Closes #363

kamicut:
After discussion with @yellowcap we were exploring the idea of creating a summary of reasoning steps and outcomes in natural language instead of a checklist.

LanesGood:
@kamicut should we fix this build and implement for now while awaiting any changes to the steps language from the backend?

yellowcap:
> @kamicut should we fix this build and implement for now while awaiting any changes to the steps language from the backend?

Yes I think that is a good idea.

LanesGood:
I've fixed the build and merged in main, so this is up to date for a latest review.

I had Claude help implement a fix for the auto-scrolling issue. The chat panel now scrolls automatically to the bottom any time a new message streams in or a state change occurs. The useEffects for scroll state are more complex now, but capture the various states and use cases; users can scroll up and expand the Reasoning component without affect the scroll position, as long as new messages have not streamed in. If the messages finish streaming, scroll navigates to the bottom of the chat panel even if a user is currently viewing the reasoning component.

@kamicut unfortunately the issue of compacting messages is still present; the bug I mentioned above still occurs: I'm going to try and parse this as well.

LanesGood:
Re-implemented the markdown and fixed the formatting issue (4+ lines of whitespace in Markdown evidently indicate a codeblock). The only other way I would recommend further collapsing this visually, in absence of an updated LLM response, would be to further nest each tool's content in a collapsible, like in this example (not pushed up):

https://github.com/user-attachments/assets/69a4ec96-cbbf-485b-b375-738d52de1bb3

However, this seems overly complex in terms of interaction. I suggest we push this forward, if approved by @01painadam , and update when the LLM response has changed.

01painadam:
@LanesGood LGTM happy to push once merge conflicts are resolved! We can add tickets to update the BE messaging to fit this template.

LanesGood (approved):
Looks good to merge @kamicut 💡 ♾️

PR #397, opened 2026-01-15
Public beta fixes

kamicut:
- generic message for unauthed page
- fix button on homepage


LanesGood:
I pushed a small zindex change here to fix the insight menu being hidden by the chat panel:

| Before | After |
| -- | --
| image | image |

PR #414, opened 2026-02-16
Configurable legend parameters & accordion UI

01painadam:
## Summary
  - Interactive legend with configurable per-layer parameters (thresholds, date
  ranges, confidence levels) — editable via popover controls that rebuild tile
  URLs on change
  - Accordion-style legend: collapse/expand layers (one at a time), auto-expands
   most recently added layer, scrollable at max height
  - Parameter context chips (threshold, confidence, date) sync to the context
  bar and attach as provenance to chart widget messages
  - Applied configurable params to 7 datasets: DIST Alerts, Land Cover,
  Grasslands, Tree Cover, Tree Cover Loss, Loss Drivers, Net Flux

Screenshot 2026-02-16 at 19 06 33

Screenshot 2026-02-16 at 19 08 40

  ## What changed
  - **New**: `ParamEditor.tsx` — popover with dual-handle range sliders, date
  pickers, categorical toggles, single-value dropdowns
  - **New**: `buildTileUrl()` / `deriveDateRange()` helpers for URL token
  replacement and human-readable date strings
  - **Modified**: `Legend.tsx`, `LayerEntry.tsx` — accordion expand/collapse,
  param editor button
  - **Modified**: `mapStore.ts` — `baseUrl`/`params` on TileLayer,
  `updateTileLayerParams` action
  - **Modified**: `contextStore.ts`, `chatStore.ts` — param-to-context sync,
  `active_params` passed to backend
  - **Modified**: `useLegendHook.tsx` — chip cleanup on layer removal, context
  snapshot on widget messages
  - **Modified**: `ContextButton.tsx`, `ContextMenu.tsx` — threshold/confidence
  chip types

Added `configurable_params` to all applicable datasets.

- **DIST Alerts (0):** `start_date`, `end_date` (date pickers), `confidence` (segmented toggle: All / High only; API values `low`/`high`)
- **Land Cover (1):** `lc_year` (segmented toggle: 2015 / 2024; only two snapshots available)
- **Grasslands (2):** `grass_start_year`, `grass_end_year` (dual-handle range slider, 2000–2022; tile uses end year for datetime filter)
- **Tree Cover (7):** `tc_threshold` (dropdown, path-based `tcd_{value}`)
- **Tree Cover Loss (4):** `start_year`, `end_year` (dual-handle range slider), `threshold` (dropdown) — already wired, updated to new types
- **Loss Drivers (8):** `threshold` (dropdown) — updated
- **Net Flux (6):** `threshold` (dropdown) — updated

  ## Checks
  - [ ] Add each of the 7 configured datasets — verify legend shows param
  controls
  - [ ] Edit threshold/date/confidence params — verify tiles reload with new URL
  - [ ] Confirm only one legend layer expanded at a time, most recent
  auto-expands
  - [ ] Remove a layer — verify its context chips are cleaned up
  - [ ] Ask a chart question with params active — verify widget message shows
  provenance chips
  - [ ] New context added in chat overrides chips

## TODO
- `projext-zeno` allow analysis using these params (e.g. TCL is currentl hardcoded to 30% threshold)


LanesGood:
Looks fantastic. A couple questions:
1. Should the layer parameters be 1:1 synced with the context chips? It appears that if I remove a parameter context chip (< 30% tree cover, for example), it does not reset/remove that parameter from the layer
2. If I reorder layers in the legend, should they reorder on the map? It does not appear that this is the case.

Suggestions:
- For the tabbed button switch option, (The confidence All/High-only option in above screenshot, and elsewhere), suggest a lighter button version.
- The scale-in transition when a layer is added to the legend is nice but a bit overdone IMO. I'd suggest a smaller transition animation. (this may be a part of the native collapsible)
- When a layer's parameters are reset to default, consider not adding these default params as context chips, if they were not there in the first place
- (For future consideration) - should/can the parameters be configurable before the user adds the layer to the map? That is, either via chat or in the context layers panel?

01painadam:
> Looks fantastic. A couple questions:
> 1. Should the layer parameters be 1:1 synced with the context chips? It appears that if I remove a parameter context chip (< 30% tree cover, for example), it does not reset/remove that parameter from the layer

Ahh thats a great point, I overlooked that. My feeling is yes, it should be bidirectional. Thoughts @kamicut @yellowcap ?

> 2. If I reorder layers in the legend, should they reorder on the map? It does not appear that this is the case.

yes, I'll implement. Good catch!

> Suggestions:

Cheers! I'll take a look and implement - thanks for the review :)

01painadam:
@yellowcap with this PR we will be able to specify confidence levels and tree cover thresholds, which should be passed to the analysis tool.

https://github.com/wri/project-zeno/issues/415

PR #420, opened 2026-03-03
Multi-AOI selection

kamicut:
* Create types for multi aoi selection
* Fetch geometries for multiple AOIs
* Add selection to context

### DIFF
--- a/app/api/shared/parse-stream-message.ts
+++ b/app/api/shared/parse-stream-message.ts
@@ -31,6 +31,7 @@ export function parseStreamMessage(
       type: "human",
       text: content as string,
       aoi: langChainMessage.aoi || undefined,
+      aoi_selection: langChainMessage.aoi_selection || undefined,
       timestamp: timestamp.toISOString(),
       start_date: langChainMessage.start_date,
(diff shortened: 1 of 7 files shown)


kamicut:
image

@yellowcap this adds a label for the bounding around the multi-area selection that can be clicked on to reselect the area

01painadam:
Looks good! I intermittently hit errors after dataset selection meaning no chart get displayed, but if I follow up it recovers nicely!

See: https://project-zeno-next-up7z3lh2q-global-nature-watch.vercel.app/app/threads/63762390-acc6-45d6-b686-4bde403a7899

One thing, we should probably have a fallback short-name based on the AOI selection e.g. States in Spain. I see in the screenshot above that it works sometimes (maybe this comes from the API?) but we it fails is there something we could do in the FE to avoid long lists?

Screenshot 2026-03-08 at 13 25 43

yellowcap:
Merging as this is a good step forward. @kamicut will work on making this layer driven, so each selection is its own layer and shows on the legend. So we are not planning to deploy this to production just yet.

PR #427, opened 2026-03-05
Env-based prototype mode

01painadam:
Adds a visual indicator for prototype deployments so testers and stakeholders can immediately distinguish them from production and avoid unrealistic assumptions about the level of quality / production-readiness

 ### What it does

 - Reads `NEXT_PUBLIC_PROTOTYPE_MODE` env var
 - When true: header turns amber, PREVIEW --> PROTOTYPE, and a centred "NOT FOR PRODUCTION USE" label appears
 - When unset: zero changes. All values fall through to defaults

gnw-proto

 ### Changes

 - app/components/PageHeader.tsx — conditional color swap + badge text + centered label via ternaries on a
 module-level isPrototype const

 ### Notes

 - No new components or layout changes
 - Only colour/text ternaries in the existing header
 - In future, prototypes and proof of concept branches will have PRs tagged as `DO NOT DEPLOY`

### DIFF
--- a/app/components/PageHeader.tsx
+++ b/app/components/PageHeader.tsx
@@ -54,8 +54,8 @@ function PageHeader() {
       px={{ base: 3, md: 5 }}
       py="2"
       h={{ base: 10, md: 12 }}
-      bg={isPrototype ? "#f59e0b" : "primary.solid"}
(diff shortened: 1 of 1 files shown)


yellowcap (approved):
I like the approach but let @kamicut chime in as well

PR #429, opened 2026-03-05
Map legends configurable params

01painadam:
Adds a ParamEditor settings popover to legend layers, allowing users to adjust dataset parameters (year, year range) directly from the legend. Parameter changes update the tile URL and legend header in real time.

Previously, dataset tile URLs were static — users couldn't change the year for land cover or adjust the date
range for tree cover loss. This wires up configurable parameters so the legend becomes interactive, not just
informational.

 ### ⚠️  Caveat: visual only

Layer setting changes are purely visual to aid exploration! Settings update the map tiles and legend header but the assistant is not aware of the current state i.e no context chips set. We need to figure out a way to make the map↔chat state sync bidirectionally. For now an approach for Tree cover threshold and DIST confidence params is also deferred to a future PR.

 ### How it works

 - ParamSpec type defines each configurable parameter: type (year, categorical), default value, URL injection
 strategy (query or path), and optional range grouping for dual-handle sliders
 - buildTileUrl() resolves path tokens (e.g. {lc_year}) and query params in tile URLs, leaving {z}/{x}/{y}
 untouched
 - ParamEditor popover renders the appropriate control per param type: segmented toggle for categorical,
 dual-handle range slider for grouped year params, single slider for standalone year. Also houses the opacity
 slider (moved from the old standalone OpacityControl)
 - addTileLayer resolves default params into the URL on initial add, so path tokens are replaced immediately
 - Legend date ranges are derived dynamically from active params instead of being hardcoded in titles — they
 update as the user changes settings

 ### Datasets configured

Dataset | Params | Control
-- | -- | --
Land cover | lc_year (2015 / 2024) | Segmented toggle
Grasslands | grass_year (2000–2022) | Year slider
Tree cover loss | start_year / end_year (2001–2024) | Dual-handle range slider

 ### Commits

 1. feat: configurable legend parameters — core infrastructure: ParamSpec type, buildTileUrl, ParamEditor
 component, opacity absorbed, default param resolution, styling polish
 2. fix: land cover year as categorical toggle — only two years available so a segmented toggle is clearer than
 a slider; initialise draft with spec defaults so the toggle shows the active value on mount
 3. fix: derive legend date ranges from params — remove hardcoded years from titles, derive dynamically from
 active params including categorical year detection


kamicut:
Hello - I'd like to pause this one until the layer manager is in and then we can go over this PR.

PR #431, opened 2026-03-12
Fix issue of accumulating tool steps

LanesGood:
This PR fixes the bug caught in #396 where reasoning steps are condensed into a single component when revisiting previous chats.

396:

https://github.com/user-attachments/assets/f15c8825-6a56-4190-adcf-3d6f24ab07f3

This PR:

https://github.com/user-attachments/assets/16c7f874-620e-4235-8363-c6ffb5b925fb

### DIFF
--- a/app/store/chatStore.ts
+++ b/app/store/chatStore.ts
@@ -51,7 +51,7 @@ interface ChatActions {
   ) => Promise<void>;
   addToolStep: (toolData: StreamMessage) => void;
   clearToolSteps: () => void;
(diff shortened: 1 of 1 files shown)


LanesGood:
I pushed this to a separate PR because it affects just the chatStore, not the actual reasoning component. This was authored with Claude, but I have a full understanding of the changes. My only concern is whether this over-prioritizes the formation of toolsteps in the chat store.

kamicut (approved):

PR #433, opened 2026-03-17
Hotfix - Remove border on reasoning component

01painadam:
Small change to remove the border while keeping the dropdown claret icon (once populated with data)

It feels out of place in the chat pane to see a big reasoning box, so trying to follow Claude-esque patterns.

New:
Screenshot 2026-03-17 at 6 03 03 PM

Any opinions @LanesGood ?

### DIFF
--- a/app/components/Reasoning.tsx
+++ b/app/components/Reasoning.tsx
@@ -65,10 +65,7 @@ function Reasoning({
         justifyContent="flex-start"
         alignItems="center"
         gap="2"
-        border="1px solid"
-        borderColor="border"
         mb={4}
-        rounded="md"
         p={2}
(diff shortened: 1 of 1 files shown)


LanesGood:
I think this looks good but defer to @faustoperez who was leading the design of what was implemented in #396.

I will also say, I think the bordered "reasoning/thinking" component was consistent with patterns from the major LLM platforms at the time, 4-6 months ago, but these have now all changed - so we have to acknowledge that if we're striving for parity with those platforms' UI/UX patterns, we'll need to make changes quite often!

faustoperez (approved):
Looks good to me :)

Issue #434, opened 2026-03-19
Layer Manager


kamicut:
Create a "Layer Manager" store that would allow us to have multiple GeoJSON layers instead of a single FeatureLayer for all the in-context features. This would improve usability around multi-area queries.

PR #435, opened 2026-03-20
Map Layer Manager

kamicut:
#434 

This adds a unified layer manager for GeoJSON and raster layers. This allows us to have multi areas as different layers on the map that can be re-ordered in the legend and interact with the raster layers as well.

This needed to be a big refactor because multiple areas in the app used to write to the map. Now they all write to a "GeoJSON registry" and then layering and rendering is separated out. @01painadam let me know if you want me to pair for the review.


01painadam:
Not blocking - but @yellowcap we should consider having the agent generate a name for the group. They're pretty long.

Screenshot 2026-03-20 at 11 44 04 AM

01painadam:
Seeing some chart gen / codeact errors (on the api side i think?)
- hhttps://project-zeno-next-i13n0ykuf-global-nature-watch.vercel.app/app/threads/1bec6ff1-ff86-4e6d-9abd-d21eeca470f3
- https://project-zeno-next-i13n0ykuf-global-nature-watch.vercel.app/app/threads/ef7b6b61-db3d-4273-b01c-c62e9c778a4c

01painadam (approved):
Nice work, LGTM! I'd consider truncating the group name a little more aggressively but not a blocker. @kamicut 

For future work we should think about how to manage areas outside of the legend - it could get packed with areas) cc: @faustoperez

PR #436, opened 2026-03-20
Add smooth scroll

01painadam:
Smoothly scroll user message to the top of the pane.

⚠️ Help needed - seems to be a small delay after hitting enter / pressing send that I cant get rid of


faustoperez (commented):
Not sure I can help with this, I'll defer to @kamicut

kamicut:
I'm not sure what you mean by delay? Do you mean the text goes from bottom to top or is there something else?

01painadam:
> I'm not sure what you mean by delay? Do you mean the text goes from bottom to top or is there something else?

When you send a prompt, there is a small time before the page actually begins to scroll

Issue #400, opened 2026-01-19
Refreshing sample prompts is broken


kamicut:
On the public landing page, prompts should "refresh" (cycle through examples) every 10 seconds. Clicking the refresh button does cycle through the list of prompts, but the timer that counts down 10 seconds does not change the prompt when it resets.

PR #439, opened 2026-03-20
Fix prompt refresh

kamicut:
Fix for #400 - landing page prompts wouldn't refresh


01painadam (approved):
LGTM - works as expected

PR #442, opened 2026-03-24
Add header and source for api calls

01painadam:
Surface clearly API docs and API sources as cards following the same pattern as code blocks and outputs

Screenshot 2026-03-24 at 12 40 06

Screenshot 2026-03-24 at 12 40 36

### DIFF
--- a/app/components/InsightProvenanceDrawer.tsx
+++ b/app/components/InsightProvenanceDrawer.tsx
@@ -13,7 +13,7 @@ import {
   IconButton,
   Button,
 } from "@chakra-ui/react";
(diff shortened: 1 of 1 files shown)


faustoperez (commented):
I suggest we show the API call details directly in whichever step makes the call, would that be possible? At the moment, it feels a bit disjointed with the summary above the steps and the details at the bottom.

01painadam:
> I suggest we show the API call details directly in whichever step makes the call, would that be possible? At the moment, it feels a bit disjointed with the summary above the steps and the details at the bottom.

it is already in the first code block its just a bit hidden... maybe we could put the api call blocks at the top _these are the sources/data we start with?_

PR #455, opened 2026-04-14
PZB-741: Removes GA4, add GTM

01painadam:
Per request from with Amanda and LauraLee:

> The GTM container includes the GA4 Property G-W7MWCJSXSF
I recommend that you install the GTM Container and remove the direct GA4 integration from the codebase and route everything through GTM instead
This will enable us to add tracking tags and custom events easily, including adding the capability to track traffic via Plausible Analytics.

Strips all GA4 event tracking (sendGAEvent/sendGAEventAsync) from the codebase and deletes the Analytics component and utility. GTM (added to `layout.tsx`) will handle analytics going forward.

Changes:
- Deleted `app/components/Analytics.tsx` and `app/utils/analytics.ts`
- Removed `<Analytics />` from root layout
- Removed 13 GA event call sites across stores, components, and pages


01painadam:
On hold while the team confirm the GTM tag

01painadam:
Confirmed - g2g for review @kamicut

Issue #444, opened 2026-03-26
Remove NextJS middleware


kamicut:
Let's remove the NextJS middleware which will allow us to create static builds and remove the amount of data transformation steps between the API and the UI.

kamicut:
Cursor Plan I will be following:

# Remove the Next.js API Proxy Layer

## Current Architecture

```mermaid
flowchart LR
  subgraph browser [Browser]
    UI[Zustand stores / Components]
  end
  subgraph nextjs ["Next.js Server (API Routes + Middleware)"]
    AuthRoutes["/api/auth/*"]
    ChatRoute["/api/chat"]
    ThreadRoutes["/api/threads/*"]
    ProxyRoute["/api/proxy/*"]
    MW[Middleware]
  end
  subgraph fastapi [FastAPI Backend]
    API["/api/chat, /api/threads, /api/metadata, ..."]
  end
  UI -->|"fetch /api/* (cookies)"| nextjs
  MW -->|"session_token, auth redirect"| nextjs
  nextjs -->|"Bearer token + X-API-KEY + X-ZENO-FORWARDED-FOR"| fastapi
```



Every browser request goes through a Next.js route handler that:

1. Reads `auth_token` / `session_token` from httpOnly cookies
2. Adds `Authorization: Bearer`, `X-API-KEY`, `X-ZENO-FORWARDED-FOR` headers
3. Forwards to the FastAPI backend
4. For chat/thread-GET: **transforms** the upstream LangChain NDJSON stream into a simplified `StreamMessage` format before forwarding to the browser

## Target Architecture

```mermaid
flowchart LR
  subgraph browser [Browser]
    UI[Zustand stores / Components]
    TokenMgr[Token Manager]
  end
  subgraph fastapi [FastAPI Backend]
    API["/api/chat, /api/threads, /api/metadata, ..."]
  end
  UI -->|"fetch direct + Bearer token"| fastapi
  TokenMgr -->|"localStorage JWT"| UI
```



The browser calls the FastAPI backend directly with `Authorization: Bearer <jwt>`. Anonymous access is dropped -- all app features require login. No Next.js server-side proxy.

---

## Key Decision: Stream Transformation

The biggest complexity is the **chat** and **thread GET** routes. They don't just proxy -- they **transform** the upstream LangChain NDJSON format into a simplified `StreamMessage` format.

**Two options:**

[...]

PR #457, opened 2026-04-16
Remove API Layer

kamicut:
- moves the authentication to direct JWT token in the client side
- client calls backend API directly with the token
- stream transformation happens on the client, not in the nextjs server side

works towards #444


kamicut:
Note to @batpad:

We need to expose these to the frontend

```
# FastAPI CORSMiddleware config
expose_headers=["X-Prompts-Quota", "X-Prompts-Used"]
```

app/dashboard/page.tsx
```diff
@@ -279,17 +282,15 @@ export default function UserSettingsPage() {
...
-      try {
-        await fetch("/api/auth/logout", { method: "POST" });
-      } catch {}
-      const url = new URL("https://api.resourcewatch.org/auth/logout");
-      url.searchParams.set("callbackUrl", `${window.location.origin}/`);
-      url.searchParams.set("origin", "gnw");
-      window.location.href = url.toString();
-    })();
+    clearToken();
+    const url = new URL("https://api.resourcewatch.org/auth/logout");
```
willian-viana: Nit: I'd add the URL to the `.env` file in case we need to use different environments (e.g. `staging-api.resourcewatch.org/...`), so we won't need to make the same change repeatedly.

willian-viana (approved):
Nice job! I only left one nit. :shipit:

PR #463, opened 2026-04-17
Fix tcl chart

solomon-negusse:
This fixes the issue of selected field for plotting in the incoming chart data being ignored.


01painadam (approved):
Nice and clean. This omits things like emissions from the polluting the TCL data as is - nice job!

PR #465, opened 2026-04-17
setup tests

guilherme-n:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

The project had no test infrastructure in place. No test runner, no coverage tooling, and no test scripts defined. This PR sets that foundation.

### Why Vitest over Jest?

Jest was my first instinct since it's the most common choice, but before going with the default I wanted to see if there was something better suited to this stack. After looking into it, Vitest felt like the right fit for a few reasons:

- Native ESM support: The codebase uses ESM modules throughout. Jest requires additional configuration and transformers (like Babel or ts-jest) to handle ESM and TypeScript, while Vitest supports both natively out of the box.
- No duplicate toolchain: Since Vitest handles its own transforms natively (TypeScript, ESM), there's no need to bolt on Babel or a separate transform layer just for tests, keeping the setup lean.
- Zero config for TypeScript: No need for ts-jest, @types/jest, or a separate Babel config — Vitest resolves TypeScript directly.
- Faster execution: Vitest leverages Vite's module graph and on-demand transforms, resulting in faster cold starts and watch-mode feedback compared to Jest.
- Compatible API: Vitest's API is largely compatible with Jest (describe, it, expect, vi vs jest), so the learning curve is minimal and switching to Jest would be straightforward if needed.
- Built-in coverage via V8: Coverage is powered by V8 natively (provider: "v8") with no extra plugins needed.
Happy to discuss if there's a preference for Jest or another approach — this is just what made the most sense to me given the current setup.

Issue Number: N/A

## What is the new behavior?

- Added **Vitest** (`^4.1.4`) as the test runner and `@vitest/coverage-v8` for coverage reporting, chosen over Jest for its native ESM/TypeScript support, zero-config setup, and faster execution in this Next.js project.
- Added vitest.config.ts with a `node` environment, V8 coverage provider, and coverage scoped to `app/**/*.{ts,tsx}` (excluding page/layout entry files and test files themselves).
- Added `test`, `test:watch`, and `test:coverage` scripts to package.json.
- Added a first test file (formatText.test.ts) covering the `toSentenceCase` utility with 9 cases (snake_case, kebab-case, spaces, mixed separators, edge cases).

## Demo
Screenshot 2026-04-17 at 13 45 12

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/app/utils/__tests__/formatText.test.ts
+++ b/app/utils/__tests__/formatText.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { toSentenceCase } from "./formatText";
+import { toSentenceCase } from "../formatText";
 
 describe("toSentenceCase", () => {
   it("converts snake_case to sentence case", () => {
--- a/app/utils/formatText.test.ts
+++ b/app/utils/formatText.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from "vitest";
+import { toSentenceCase } from "./formatText";
+
+describe("toSentenceCase", () => {
+  it("converts snake_case to sentence case", () => {
+    expect(toSentenceCase("hello_world")).toBe("Hello World");
+  });
+
+  it("converts kebab-case to sentence case", () => {
+    expect(toSentenceCase("tree-cover-loss")).toBe("Tree Cover Loss");
+  });
+
+  it("converts space-separated words", () => {
+    expect(toSentenceCase("hello world")).toBe("Hello World");
+  });
+
+  it("handles already capitalized words", () => {
+    expect(toSentenceCase("Hello World")).toBe("Hello World");
+  });
+
+  it("handles mixed separators", () => {
+    expect(toSentenceCase("area_km2-value")).toBe("Area Km2 Value");
+  });
+
+  it("handles a single word", () => {
+    expect(toSentenceCase("forest")).toBe("Forest");
+  });
+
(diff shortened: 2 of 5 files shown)


willian-viana:
I like the choice of Vitest. It's lightweight and has been supported by Next.js since version 14. 

Regarding the structure, I'd go with a `__tests__` folder instead of placing tests at the same level of the `.ts` files.

guilherme-n:
> I like the choice of Vitest. It's lightweight and has been supported by Next.js since version 14.
> 
> Regarding the structure, I'd go with a `__tests__` folder instead of placing tests at the same level of the `.ts` files.

Thanks for the feedback! I considered `__tests__` but prefer co-location for a few reasons:

- Discoverability: in a project with low coverage, seeing `formatText.ts` without a `formatText.test.ts` right next to it makes gaps immediately obvious.
- Refactoring: when moving or renaming a file, the test moves with it automatically; with `__tests__` it's easy for them to drift out of sync.
- Cognitive load: you don't need to mentally map between two folder trees to find the test for a given file.

The `__tests__` convention makes more sense in projects with high coverage where tests are the norm. Here it would hide the gaps.

PR #466, opened 2026-04-20
Fix bbox for antimeridian crossings

01painadam:
- Adds bbox: `[number, number, number, number]` to the AOI type the backend already provides this field in `[west, south, east, north]` format where west > east signals antimeridian crossing          
- Adds flyToBounds action to mapStore that normalises antimeridian-crossing bounds before calling fitBounds (east += 360 when west > east, which MapLibre handles correctly)                  
- Updates pickAoi to use backend-provided bbox values to compute combined fly-to bounds instead of Turf's bbox (which returns [-180, s, 180, n] for antimeridian geometries)                           
- Updates GeoJsonLayers bbox overlay to prefer backend bbox values with the same normalisation, falling back to Turf for layers without a backend bbox (e.g. custom drawn areas)

| Before | After |
| --- | --- |
| Screenshot 2026-04-20 at 11 25 02 | Screenshot 2026-04-20 at 11 23 48 |

This PR **MUST ship** before the backend fix `project-zeno` [fix/antimeridian-bbox](https://github.com/wri/project-zeno/pull/630) both the fly-to and overlay paths fall back to the existing Turf-based behaviour when aoi.bbox is absent or a plain non-crossing bbox.

### DIFF
--- a/app/store/chat-tools/pickAoi.ts
+++ b/app/store/chat-tools/pickAoi.ts
@@ -3,7 +3,6 @@ import { ChatMessage, StreamMessage, AOI, AOISelection } from "@/app/types/chat"
 import useMapStore from "../mapStore";
 import useContextStore from "../contextStore";
 import { fetchGeometry } from "@/app/utils/geometryClient";
-import bbox from "@turf/bbox";
 import { GeoJsonEntry } from "../layerManagerSlice";
 import { selectLayerOptions } from "@/app/types/map";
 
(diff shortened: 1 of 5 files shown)


app/store/chat-tools/pickAoi.ts
```diff
@@ -134,41 +134,30 @@ export async function pickAoiTool(
...
-      let minX = Infinity, minY = Infinity, maxX = -Infinity, maxY = -Infinity;
-      for (const geoData of allGeoData) {
-        const b = bbox(geoData);
-        if (b[0] < minX) minX = b[0];
-        if (b[1] < minY) minY = b[1];
-        if (b[2] > maxX) maxX = b[2];
-        if (b[3] > maxY) maxY = b[3];
+    console.log("[pickAoi] successfulAois bbox check:", successfulAois.map((a) => ({ name: a.name, bbox: a.bbox })));
+
+    // Fly to the combined bounds using backend-provided bbox values.
```
jterry64, line 145: Oh, disregard my comment about whether this bbox was used on the other review. However, the bbox in the other PR is per AOI - here you need to combine all bboxes to get an encompassing, which could also pass the meridian, right?

jterry64 (changes requested):

01painadam:
@jterry64 Now has a fix for antimeridian crossing with in multiaoi queries by unwrapping individual aoi bboxes.

Extracts the duplicated bbox union loop from pickAoi.ts and GeoJsonLayers.tsx into a shared bboxUtils.ts utility, then updated both callers to use it.

PR #483, opened 2026-05-12
PZB-896 - Redirect to / and extract to handler

01painadam:
## Summary
Fixes logout redirect issue.

- Wire the `useLogout` hook into the three logout call sites (sidebar, PageHeader, dashboard) so logout redirects to `/`instead of the RW logout URL.
- Removes the duplicated inline logout logic that previously lived in each component.

### DIFF
--- a/app/hooks/useLogout.ts
+++ b/app/hooks/useLogout.ts
@@ -1,6 +1,7 @@
 import { useState } from "react";
 import { toaster } from "@/app/components/ui/toaster";
 import { clearToken } from "@/app/lib/api-client";
+import { API_CONFIG } from "@/app/config/api";
 
 export function useLogout() {
   const [isLoggingOut, setIsLoggingOut] = useState(false);
@@ -15,9 +16,14 @@ export function useLogout() {
         type: "info",
         duration: 8000,
       });
-    } catch {}
+    } catch (error) {
+      console.error("Failed to show logout toast:", error);
+    }
     clearToken();
-    window.location.href = "/";
+    const url = new URL(`${API_CONFIG.RW_API_HOST}/auth/logout`);
+    url.searchParams.set("callbackUrl", `${window.location.origin}/`);
+    url.searchParams.set("origin", "gnw");
+    window.location.href = url.toString();
   };
 
   return { logout, isLoggingOut };
--- a/app/hooks/useLogout.ts
+++ b/app/hooks/useLogout.ts
@@ -1,0 +1,24 @@
+import { useState } from "react";
+import { toaster } from "@/app/components/ui/toaster";
+import { clearToken } from "@/app/lib/api-client";
+
+export function useLogout() {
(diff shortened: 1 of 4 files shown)


app/hooks/useLogout.ts
```diff
@@ -0,0 +1,24 @@
...
+        type: "info",
+        duration: 8000,
+      });
+    } catch {}
+    clearToken();
+    window.location.href = "/";
+  };
+
+  return { logout, isLoggingOut };
+}
```
guilherme-n, line 30: Nice job by extracting this code into a hook, but I noticed you removed the code that ends the session on the server. Is that expected?
```ts
const url = new URL(`${API_CONFIG.RW_API_HOST}/auth/logout`);
url.searchParams.set("callbackUrl", `${window.location.origin}/`);
url.searchParams.set("origin", "gnw");
window.location.href = url.toString();
```

app/hooks/useLogout.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (isLoggingOut) return;
+    setIsLoggingOut(true);
+    try {
+      toaster.create({
+        title: "Logging out",
+        description: "Signing you out and redirecting…",
+        type: "info",
+        duration: 8000,
+      });
+    } catch {}
```
guilherme-n: maybe add a `console.error` to prevent silent errors in the `catch` and help debugging on prod in case of failure in the toaster

app/hooks/useLogout.ts
```diff
@@ -0,0 +1,24 @@
+import { useState } from "react";
+import { toaster } from "@/app/components/ui/toaster";
+import { clearToken } from "@/app/lib/api-client";
+
+export function useLogout() {
+  const [isLoggingOut, setIsLoggingOut] = useState(false);
```
guilherme-n, line 7: small improvement suggestion: once `window.location.href` is set, navigation starts and the component unmounts. If you use a `ref`, it avoids an unnecessary re-render. Something like this (in case the goal is to prevent double clicks).
```ts
const isLoggingOut = useRef(false);

const logout = () => {
  if (isLoggingOut.current) return;
  isLoggingOut.current = true;
  // ...
};
```

app/hooks/useLogout.ts
01painadam: Oh, nice catch - threw the baby out with the bathwater!

app/hooks/useLogout.ts
01painadam: Good idea. One thing to flag: isLoggingOut is also consumed in app/sidebar.tsx:328-329 to drive loading/disabled on the mobile logout button, so we can't go pure-ref without losing that UI feedback.

I'd suggest leaving as is for now

guilherme-n: That makes sense :)

guilherme-n (approved):

PR #485, opened 2026-05-14
PZB-841: update dataset card design

guilherme-n:
## Pull request type

 

 

- [x] Feature

DatasetCard was a selectable/clickable card with a dynamic border that highlighted on selection and hover, showing dataset.reason || dataset.description as the subtitle. The DatasetCardConfig type had no metadata fields (cadence, resolution, geographic coverage, provider).

## What is the new behavior?
DatasetCard is now a compact, non-interactive display card with fixed dimensions (height: 5rem, image width: 5rem) and a subtle static border (1px solid neutral.900/30); the selected, onClick props and hover/selection styles have been removed
The InfoIcon is now rendered inline inside the card title (alongside the dataset name) instead of as an absolute-positioned button, and the subtitle is hardcoded to "Yearly · Global · WRI"
DatasetCardConfig now includes four new metadata fields — cadence, resolution, geographic_coverage, and provider — populated for all existing dataset entries

- [ ] Yes
- [x] No

## Demo
Screenshot 2026-05-14 at 17 18 15

### DIFF
--- a/app/components/ContextMenu.tsx
+++ b/app/components/ContextMenu.tsx
@@ -1,4 +1,5 @@
 import { Suspense, useEffect, useMemo, useState } from "react";
+import type { DatasetInfo } from "@/app/types/chat";
 import {
   Box,
   Card,
@@ -29,7 +30,7 @@ const CONTEXT_NAV = (Object.keys(ChatContextOptions) as ChatContextType[]).map(
   })
 );
 
-import { DATASET_CARDS } from "../constants/datasets";
+import { DATASET_CARDS, DatasetCardConfig } from "../constants/datasets";
 import { useCustomAreasListSuspense } from "../hooks/useCustomAreasList";
 import type { CustomArea } from "../schemas/api/custom_areas/get";
 import useMapStore from "../store/mapStore";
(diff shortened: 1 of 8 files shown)


app/components/DatasetCard.tsx
01painadam, line 1: Just needs to be aligned with [the designs](https://www.figma.com/design/a4jBBuwr0FQfKTY3l8aOKM/-Global-Nature-Watch--Phase-3?node-id=110-523&t=r74gvdlUdtoqfala-4)

Target:
Image

app/components/ContextMenu.tsx
01painadam, line 1: I realised while testing that in the context menu we need to retain this interaction. It just needs to be disabled in the chat context

Sorry, I totally overlooked the fact that these are shared components.

01painadam:
@guilherme-n made a few changes - please review:
- Added tooltip - this is a slightly new pattern so made it resuable as variant="dark"
- Injected the reasoning message after the widget message in `pickDataset`
- Reapplied the interaction only in ContextMenu so the cards can be used to select datasets as context
- Applied the styles from the design using the figma MCP 🤖 (and added a consistent buffer above/below the card)

01painadam (approved):

PR #487, opened 2026-05-14
PZB-897: Frontend for Manage Users screen for superusers to change users type

batpad:
This:

 - Adds a new `/manage-users/` route where superusers can search for users and modify their "user type"
 - Adds a link to Manage Users on the `dashboard` page if a user is `superuser`
 - On the Manage Users page, presents a UI to search for a user, show results, and then modify the user type of a user in the result list.

This is a demo of what it looks like:

https://github.com/user-attachments/assets/73e76e6e-0dc7-4a16-82c1-2e5678acf1c2

@kamicut I think I'm mostly concerned about the refactoring that pulled out stuff from the dashboard to re-use on the Manage Users page. I tried to split up the steps into separate commits to make reviewing easier, but let me know if there's anything else I can do?

To test: You need to be a `superuser` on staging. @kamicut I have made your devseed email a superuser. If anyone else is testing this, let me know and I can bump you to superuser on staging.

### DIFF
--- a/app/components/SettingsShell.tsx
+++ b/app/components/SettingsShell.tsx
@@ -1,0 +1,149 @@
+"use client";
+import { ReactNode } from "react";
+import {
+  Box,
+  Button,
+  ButtonGroup,
+  Flex,
+  Heading,
+  Progress,
+  Text,
+  Link as ChakraLink,
+} from "@chakra-ui/react";
+import {
+  GearIcon,
+  LifebuoyIcon,
+  MapTrifoldIcon,
+  SignOutIcon,
+  UserIcon,
+  UsersThreeIcon,
+} from "@phosphor-icons/react";
+import Link from "next/link";
+import LclLogo from "@/app/components/LclLogo";
+import useAuthStore from "@/app/store/authStore";
+import { useLogout } from "@/app/hooks/useLogout";
+
+type SettingsPath = "/dashboard" | "/manage-users";
+
+interface SettingsShellProps {
+  activePath: SettingsPath;
+  children: ReactNode;
+}
+
+export default function SettingsShell({
+  activePath,
+  children,
+}: SettingsShellProps) {
+  const { userEmail, userType, usedPrompts, totalPrompts } = useAuthStore();
+  const { logout } = useLogout();
+
+  return (
+    <Box
+      display="grid"
+      gridTemplateColumns="20rem 1fr"
+      height="100vh"
+      maxH="100vh"
+    >
+      <Flex flexDir="column" bg="bg.subtle" px={6} py={8} maxH="100%" gap={6}>
+        <ChakraLink
+          as={Link}
+          href="/"
+          display="flex"
+          alignItems="center"
+          gap="2"
+          transition="opacity 0.24s ease"
+          _hover={{ opacity: 0.8, textDecor: "none" }}
+        >
+          <LclLogo width={16} avatarOnly />
+          <Heading m={0}>Global Nature Watch</Heading>
+        </ChakraLink>
+        <ButtonGroup
+          size="sm"
+          w="full"
+          gap={2}
+          variant="outline"
+          _hover={{ "& > :not(:hover)": { opacity: "0.5" } }}
+          css={{ "& > *": { justifyContent: "flex-start", width: "100%" } }}
+          colorPalette="gray"
+          orientation="vertical"
+          alignItems="stretch"
+        >
+          <Button
+            asChild
+            bg={activePath === "/dashboard" ? "white" : undefined}
+          >
+            <Link href="/dashboard">
+              <GearIcon />
+              User Settings
+            </Link>
+          </Button>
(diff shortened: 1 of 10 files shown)


batpad:
Added a Download CSV button as well to allow superusers to download a CSV of all users - calls the backend endpoint created in https://github.com/wri/project-zeno/pull/670

app/manage-users/page.tsx
guilherme-n, line 1: I noticed this new page is hidden for non superusers with `{userType === "superuser" && (`, but if someone tries to access the url directly, then the page would be accessed by non superusers. I believe it would be good to block on server side too, maybe something like:
```ts
useEffect(() => {
  if (userType !== "superuser") {
    // redirect to main page
  }
}, [userType]);
```

app/manage-users/page.tsx
```diff
@@ -0,0 +1,270 @@
...
+      }),
+    [isSelf]
+  );
+
+  return (
+    <Select.Root
+      collection={collection}
+      size="sm"
+      width="180px"
+      value={[user.userType]}
```
kamicut, line 63: This label will be undefined for a machine user.

app/components/SettingsShell.tsx
```diff
@@ -0,0 +1,149 @@
...
+          variant="outline"
+          _hover={{ "& > :not(:hover)": { opacity: "0.5" } }}
+          css={{ "& > *": { justifyContent: "flex-start", width: "100%" } }}
+          colorPalette="gray"
+          orientation="vertical"
+          alignItems="stretch"
+        >
+          <Button
+            asChild
+            bg={activePath === "/dashboard" ? "white" : undefined}
```
kamicut: Can you use a color token from the theme? In case of dark mode this will be hard-coded.

app/store/authStore.ts
```diff
@@ -1,9 +1,12 @@
 import { create } from "zustand";
 import { API_CONFIG } from "@/app/config/api";
 
+export type UserType = "regular" | "admin" | "pro" | "superuser" | "machine";
+
```
guilherme-n: those types are already defined at `app/schemas/api/admin/users/get.ts`. What about re-exporting them so that they dont get out of sync? something like this:
```ts
import { UserTypeEnum } from "@/app/schemas/api/admin/users/get";
export type UserType = z.infer<typeof UserTypeEnum>;
```

app/manage-users/page.tsx
```diff
@@ -0,0 +1,270 @@
...
+import useAuthStore from "@/app/store/authStore";
+import SettingsShell from "@/app/components/SettingsShell";
+import { useAdminUsersList } from "@/app/hooks/useAdminUsersList";
+import { useAdminUserTypeUpdate } from "@/app/hooks/useAdminUserTypeUpdate";
+import { useAdminUsersExport } from "@/app/hooks/useAdminUsersExport";
+import { toaster } from "@/app/components/ui/toaster";
+import { type UserModel } from "@/app/schemas/api/admin/users/get";
+import { type AssignableUserType } from "@/app/schemas/api/admin/users/patch";
+
+const ASSIGNABLE_TYPES: AssignableUserType[] = [
```
kamicut: Can you reuse the enum from the users schema?

guilherme-n (approved):

app/store/authStore.ts
```diff
@@ -1,9 +1,12 @@
 import { create } from "zustand";
 import { API_CONFIG } from "@/app/config/api";
 
+export type UserType = "regular" | "admin" | "pro" | "superuser" | "machine";
```
kamicut: This is also defined in the API schema

app/manage-users/page.tsx
batpad: It's fine if someone lands up accessing the page, they won't be able to perform any actions or see any user information because any of the data coming from the backend requires a token, I don't see this as a concern.

kamicut (commented):
All looks good and functional to me! There are a couple of type duplications and minor style things that I think could warrant a fix. I also noticed that as a regular user I could navigate directly to `manage-users`. It won't work but maybe we should add a client only <Unauthorized> Component for that case.

app/store/authStore.ts
batpad: thanks for the catch! will do!

app/manage-users/page.tsx
batpad: Can you explain that a bit to me?

batpad:
Thanks much for the review and comments @guilherme-n @kamicut !

I've addressed the feedback - replaced the duplicated type definitions, and replaced the hard-coded colour.

Minor thing: when importing AssignableUserType, I see that the user types are defined as:
 - regular
 - admin
 - pro
 - superuser

I would like this to be:
 - regular
 - pro
 - admin
 - superuser

(pro before admin to follow the hierarchy of types)

I can make that minor change in a separate PR after double-checking that it won't have any unintended side-effects.

PR #489, opened 2026-05-18
feat(claude): add CLAUDE.md

willian-viana:
Adds a CLAUDE.md for working on this project. The goal is to avoid Claude Code from starting fresh every time and having to guess the commands and project workflow. No behavioral instructions were added since it's a new team and a new project (at least for @guilherme-n and me) and we haven't faced issues with recurring prompt instruction yet. 

Please, feel free to suggest any addition or removals.

PS. I've added commit conventions as a suggestion only. I still see value in using conventional commits nowadays.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,159 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commands
+
+```bash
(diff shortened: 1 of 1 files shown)


01painadam (approved):
I like starting lean and focusing on context - smart approach!

One suggestion would be adding something to capture big decisions as ADRs in a docs folder. Not something we need now, but might help us in the future (and help updating the CLAUDE.md as the project context evolves)

Nice work!

PR #490, opened 2026-05-19
[PZB-840] Legend redesign

kamicut:
This redesigns the legend to match the new Figma designs:

1. The legend is moved to a new collapsible component with a max-height
2. The layer cards now have Params chips
3. The context_layer is now nested within the parent layer
4. AOIs show up as chips instead of separate vector layers

I kept the possibility for having geojson / vector layers in the future, they have simple legend cards

@faustoperez there are some colors from your designs I used that were not in the saved variables in our theme, so I have to review this with you to see if these are supposed to be theme colors or not.

I also fixed what I think is a bug on staging: adding multiple administrative areas will not add them all in the context, it will add the last selection.

Please try out the preview and let me know if something is missing. I know this is a big PR so I can pair with whoever will review it.

### DIFF
--- a/app/components/ContextMenu.tsx
+++ b/app/components/ContextMenu.tsx
@@ -242,7 +242,7 @@ function AreaCardList({
 function AreaMenu() {
   const { customAreas } = useCustomAreasListSuspense();
   const { addToRegistry, addLayer, flyToGeoJsonWithRetry } = useMapStore();
-  const { context, upsertContextByType } = useContextStore();
+  const { context, addContext } = useContextStore();
   const [query, setQuery] = useState("");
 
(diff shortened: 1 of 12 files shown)


01painadam (changes requested):
Looks good so far! One minor bug noted when adding areas via click-on-map interaction to take care

01painadam (approved):
Bug fixed, all works as intended. Send it!

faustoperez:
Looking good @kamicut !

I identified a couple of things we should review together:
- Expand / collapse icons
- Title behavior when there's multiple lines
- How categories are displayed
- Consolidate colors with the system colors

image

PR #494, opened 2026-05-21
PZB-910: Fix missing axis specification error

yellowcap:
## Summary

  - Relax the axis guard from !xAxis || !yAxis to !xAxis || (type === "scatter" && !yAxis) — scatter charts require an explicit yAxis; all other types can derive series automatically from data columns when yAxis is empty
  - Apply the same fix to the useMemo gate so formatChartData is actually called for multi-series charts instead of returning { data: [], series: [] } immediately

## Context

The backend uses yAxis="" for multi-series bar/line/area charts (wide format), populating seriesFields instead. formatChartData already handles this correctly — when yAxis is empty it treats every non-x column as a series — but the xAxis && yAxis guard silently bypassed it, causing "Chart is missing axis configuration." to appear whenever the AI produced a multi-series chart.

  The backend has a paired fix (in project-zeno) that pre-filters each chart's data to only its relevant columns, so formatChartData's column auto-discovery produces the right series count.

This also needs some updates on the backend in parallel.


01painadam:
Merged develop, no conflicts with the ChartWidget change.

Two things worth a look before/after the parallel backend change lands:

1. **Pie charts will silently render empty when `yAxis` is missing.** The relaxed guard lets pie reach the renderer, but `ChartWidget.tsx:340` still passes `dataKey={chart.key(yAxis)}` directly. `formatChartData` computes an internal `valueKey` fallback (`formatCharts.tsx:79`) but it isn't propagated to the component. internal `valueKey` fallback (`formatCharts.tsx:79`) but it isn't propagated to the component. Either extend the strict guard to include `type === "pie" && !yAxis`, or return the resolved `valueKey` from `formatChartData` so pie can use it.

2. **Type lies about `yAxis`.** `InsightWidget.yAxis: string` is declared required (`types/chat.ts:45`), but the component now treats it as possibly undefined. Should be `yAxis?: string`.

yellowcap:
Depends on https://github.com/wri/project-zeno/pull/681

yellowcap:
With latest changes in https://github.com/wri/project-zeno/pull/681 this yielded nice results. But I can't review the react code properly so that definitively needs a good review.

| # | Query | Screenshot |
|---|--------|------------|
| 1 | Plot a timeseries of tree cover loss comparing Indonesia and Malaysia |  |
| 2 | Compare land cover distribution between the two countries — how much natural land does each have? |  |
| 3 | Make the previous TCL data a multi-line plot |  |

01painadam (approved):
Nice fix, works for me!

PR #497, opened 2026-05-21
Map insights tweaks

01painadam:
Insight workspace + legend Figma alignment, chat-side analysis cards

- Visual alignment: InsightWorkspace and Legend share a 420px panel
with the same 28px Figma section header (dark "AI-Assisted Analysis ·
learn more" tooltip, type-aware icon, #DDE2F5 dividers, #F7F9FF bg).
AREA chips collapsed to one per selection (matches the legend) and the
AREA value text now reads in the same blue as its label.
- Chat-side analysis cards: Each insight reference message in chat
renders one AnalysisCard per chart widget (clickable surface for
future workspace navigation). The thumbs-up/down/copy/timestamp footer
is now forced on for insight references and the final message in any
conversation (previously hidden by the isConsecutive rule).
- Workspace UX nits: Prev/Next direction flipped so 1/N disables the
left arrow and the right arrow steps through to older entries. "Show
params" toggle uses the same #4A64CB blue as the header "learn more"
link, and the older-conversations sidebar section is collapsed by
default.

### DIFF
--- a/app/components/AnalysisCard.tsx
+++ b/app/components/AnalysisCard.tsx
@@ -1,0 +1,59 @@
+"use client";
+import { InsightWidget } from "@/app/types/chat";
+import { WidgetIconComponent } from "@/app/utils/widgetIcons";
+import { InfoCard } from "./InfoCard";
+
+const CHART_TYPE_LABEL: Record<InsightWidget["type"], string> = {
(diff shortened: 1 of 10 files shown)


willian-viana (approved):
:shipit:

PR #500, opened 2026-05-25
Tidy insight workspace, legend layout, and Insight Card chat injection

01painadam:
### Summary
Worked with Fausto on friday to review the behaviour of the Insight+Legends. A couple of things tidied up in this PR:

- Insight workspace + legend no longer overlap. A single shared right-column flex container owns the 420 px right edge of the map. InsightWorkspace sits at the top (content-sized, can shrink), a spacer pushes the legend to the bottom, and the two panels always maintain a gap — regardless of chart height.
- InsightWorkspace scrolls correctly. The card is its own scroll container (overflowY="auto"). The "AI-ASSISTED ANALYSIS" header and the "N of M available analyses" navigation footer are position="sticky" so they never scroll out of view. Only the chart body scrolls.
- Map tools anchored to the chat panel (left). MapAreaControls was centred horizontally across the map, making it vulnerable to being obscured by the insight panel. Changed to alignItems="flex-start" so tools always sit top-left, clear of the right-column stack.
- Insight cards injected inline in chat. The previous behaviour added a static "I've created an insight you can view on the map." message when the tool call completed, then the LLM text referenced the chart again as [Chart <uuid>]. The static message is gone. Instead, when the assistant's text message arrives in the stream, any [Chart <uuid>] markers are split out and replaced with the corresponding AnalysisCard widget, injected positionally from insightStore.pendingBatch. Text segments preceding a card get tight spacing and no footer controls since they flow directly into the chart.

Test plan
Generate chart e.g. "Trends in Natural Grassland extent in Mongolia"
- Confirm the chart card appears inline in the assistant message at the point the LLM references it — no preceding "I've created an insight" message.
- Ask a question that generates multiple insights (e.g. TCL in BRA). Confirm each [Chart …] reference is replaced by the correct chart in order.
- With an insight visible, confirm InsightWorkspace (top-right) and Legend (bottom-right) are separated by a gap and do not overlap.
- Resize the browser to make the insight panel tall enough to need scrolling. Confirm the header and "N of M" footer stay fixed while the chart body scrolls.
- Collapse the insight panel using the caret. Confirm only the header remains and the legend moves up to fill available space.
- Use the upload / select / draw tools. Confirm they appear at the top-left of the map and are not obscured by the insight or legend panels.
- Load a historical thread that contains insights e.g. `http://localhost:3000/app/threads/c10551f5-c183-448c-af89-cb8f7d10ff75` Confirm inline cards render correctly on replay.

### DIFF
--- a/app/components/InsightWorkspace.tsx
+++ b/app/components/InsightWorkspace.tsx
@@ -66,37 +66,35 @@ export default function InsightWorkspace() {
   const HeaderIcon = WidgetIconComponent[widget.type];
 
   return (
+    /* Card is the scroll container: grows to content, scrolls when flex-shrunk */
     <Box
-      position="absolute"
-      top={4}
-      right={3}
-      zIndex={400}
-      pointerEvents="none"
+      flex="0 1 auto"
+      minH="0"
+      overflowY="auto"
+      w="100%"
+      bg="primary.25"
+      border="1px solid"
+      borderColor="#DDE2F5"
+      rounded="4px"
+      pointerEvents="all"
+      display="flex"
+      flexDirection="column"
     >
-      {/* Panel */}
-      <Box
-        w="420px"
-        maxH="calc(100vh - 6rem)"
-        overflowY="auto"
+      {/* Header — sticky so it never scrolls away */}
+      <Flex
+        position="sticky"
+        top={0}
+        zIndex={1}
         bg="primary.25"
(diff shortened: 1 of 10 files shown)


faustoperez (approved):
Works as expected, looks good to me :)

guilherme-n (approved):

Issue #440, opened 2026-03-24
Paginate Threads pull


01painadam:

batpad:
Am happy to take this on as it will require changes on both the backend and frontend.

To just try and outline the problem a bit clearly to ensure I understand things correctly:

Currently, the list of threads populated in the left side-bar can get really long - right now, the entire list of threads is fetched as a single request from the backend, without any pagination - this will not scale for power users who could have hundreds of threads.

**Proposed Solution**: For the UX, I'd propose a simple "infinite scroll" behaviour for users - so no explicit pagination controls, but we just load more past threads as the user scrolls down in the list. This seems like the most straightforward for users and avoids complexity of adding pagination controls, etc.

Looking at it, am tempted to also propose a simple "search" feature to allow users to search for past threads - tho likely searching through anything more than the title can likely get complicated. We can worry about this later.

**Implementation**: On the backend, implement some pagination mechanism via query parameters - likely `limit` and `offset` query parameters for the frontend to fetch items incrementally, and implement a `max_limit` of 100 items or so to prevent the possibility of extremely large requests. On the frontend, implement "infinite scroll" behaviour that fetches more items as the user scrolls down - with some loading indicator to show items being fetched and likely some good way to indicate that the list is complete / there is nothing more to fetch.

I'll try and plugin some default infinite scroll behaviour and then we can probably iterate on polishing design.

Let me know if there's some infinite scroll plugin that we like and that I should use? cc @kamicut 

cc @yellowcap @01painadam

PR #501, opened 2026-05-25
implement infinite scroll pagination for conversations / threads listing

batpad:
Refs #440 

Depends on https://github.com/wri/project-zeno/pull/685 backend PR, although nothing should break if deployed without the backend change.

Implements "Infinite scrolling" for threads - fetching upto , and then fetching more if the user scrolls down.

Am not fully sure what's going on with this infinite scroll plugin - this was definitely quite vibe-coded, so would appreciate any feedback.

To test:

 - Needs to point to new version of backend with https://github.com/wri/project-zeno-next/issues/440 merged
 - If you have less than 100 threads, adjust PAGE_SIZE in `app/hooks/useThreadsInfinite.ts` to make it smaller
 - Should fetch more threads as you scroll down, and not all threads in a single request

This will probably be easier to test once the backend is merged to staging.

### DIFF
--- a/app/components/PageHeader.tsx
+++ b/app/components/PageHeader.tsx
@@ -34,6 +34,7 @@ import useSidebarStore from "../store/sidebarStore";
 import ThreadActionsMenu from "./ThreadActionsMenu";
 import Link from "next/link";
 import { useLogout } from "@/app/hooks/useLogout";
+import { useThreadsInfinite } from "@/app/hooks/useThreadsInfinite";
 
 const isPrototype = process.env.NEXT_PUBLIC_PROTOTYPE_MODE === "true";
 const DISCLAIMER_STORAGE_KEY = "gnw_disclaimer_dismissed_v2";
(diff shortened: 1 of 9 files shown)


01painadam:
@batpad one small this - the loading should check if the section is open first. Since `Older Conversations' starts closed by default I wouldn't expect it to start loading until it is opened

batpad:
@01painadam it loads the first 100 threads by default - so Older Conversations would have some loaded, but then should only load further as you scroll down? If it does seem strange to you, do you mind recording a quick screen-cast?

01painadam:
@batpad the loading flickering while the section is closed (I imagine it _thinks_ it needs to keep loading all the pages)
load-ezgif com-cut

batpad:
Thanks much for that catch @01painadam . I've pushed a fix that should make the loading behaviour better. Could you give it another test?

batpad:
Rebased this on develop. @01painadam @kamicut mind giving this a look, would be nice to get this merged in before the next release.

PR #507, opened 2026-05-28
PZB-913: Implement the fires dataset nudge

willian-viana:
The agent picks a dataset silently based on a user's prompt even if there are other possible options.

## What is the new behavior?
Given the user's prompt is fires-related and more than one fire dataset could plausibly answer it, when the agent replies, then it surfaces those candidate datasets instead of picking one silently.

- [ ] Yes
- [X] No

### DIFF
--- a/app/components/DatasetNudge.tsx
+++ b/app/components/DatasetNudge.tsx
@@ -8,7 +8,7 @@ import useContextStore from "@/app/store/contextStore";
 import { DATASET_BY_ID } from "@/app/constants/datasets";
 import { getDatasetLayerContextProps } from "@/app/utils/datasetLayerContext";
 
-export default function DatasetNudgeWidget({
+export default function DatasetNudge({
   datasets,
 }: {
   datasets: SuggestedDataset[];
--- a/app/components/DatasetNudge.tsx
+++ b/app/components/DatasetNudge.tsx
@@ -19,10 +19,10 @@ export default function DatasetNudge({
     if (pickedId !== null) return;
     setPickedId(selected.dataset_id);
 
-    const localDataset = DATASET_BY_ID[selected.dataset_id];
-    if (localDataset) {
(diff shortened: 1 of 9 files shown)


gtempus:
Looks great, @willian-viana ! 💪

app/components/DatasetNudge.tsx
```diff
@@ -0,0 +1,120 @@
...
+        content: merged.dataset_name,
+        datasetId: merged.dataset_id,
+        tileUrl: merged.tile_url,
+        layerName: merged.dataset_name,
+        ...layerContextProps,
+        isAiContext: true,
+      });
+    }
+
+    useChatStore.getState().sendMessage(selected.dataset_name, "human_input");
```
guilherme-n, line 46: Should this call be inside the `if (localDataset)` block, or is it truly independent of it?

guilherme-n (approved):

app/types/chat.ts
```diff
@@ -157,6 +166,17 @@ export interface DatasetParameter {
...
 
+export interface SuggestedDataset {
+  dataset_id: number;
+  dataset_name: string;
+  context_layer?: string | null;
+  parameters?: DatasetParameter[] | null;
+  start_date?: string;
+  end_date?: string;
+  reason?: string;
+  recommended?: boolean;
```
jterry64, line 177: Ooo I don't think I added the recommendation field yet, I can try to add that today, but it looks like it's nullable at least.

jterry64 (commented):
I won't leave a review because I don't know this code well at all, but I loaded it up locally and it works great for me!

app/types/chat.ts
willian-viana: Yes, it's optional for now

app/components/DatasetNudge.tsx
willian-viana: Yes, it should be independent.

gtempus (approved):

PR #509, opened 2026-06-01
PZB-839:  Move chat controls to header

01painadam:
Move chat controls from chat to the header and aligns styles with upcoming changes.

Screenshot 2026-06-01 at 13 06 23

- Moves Conversation History, Conversation Name + sharing, New conversation to header
- Updated header styles (retaining Blue primary colour)
- Removes "Go to Insights" dropdown (and tidies up)
- Adds new chat `AI ASSISTANT` header aligned with upcoming chat pane changes

### DIFF
--- a/app/ChatPanelHeader.tsx
+++ b/app/ChatPanelHeader.tsx
@@ -1,7 +1,6 @@
-import { useCallback, useMemo } from "react";
-import { Flex, Menu, Button, Portal, Text } from "@chakra-ui/react";
+import { Flex, Text } from "@chakra-ui/react";
 import {
-  CaretDownIcon,
+  SparkleIcon,
   ChartLineIcon,
   ListNumbersIcon,
   ChartBarIcon,
@@ -11,9 +10,6 @@ import {
   ChartScatterIcon,
   ChartPolarIcon,
 } from "@phosphor-icons/react";
-
-import { Tooltip } from "./components/ui/tooltip";
-import useChatStore from "./store/chatStore";
 
 export const WidgetIcons = {
   line: <ChartLineIcon />,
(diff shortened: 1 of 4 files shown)


app/ChatPanelHeader.tsx
```diff
@@ -1,271 +1,32 @@
...
-  }, []);
-
   return (
     <Flex
+      h="40px"
+      mt="3"
+      mx="3"
+      px="3"
+      py="1"
+      bg="#f4f5f6"
```
guilherme-n: do you think we could get this color from the chakra theme?

app/ChatPanelHeader.tsx
```diff
@@ -1,271 +1,32 @@
...
-      )}
+      <SparkleIcon size={16} color="#656e7b" />
+      <Text
+        fontFamily="mono"
+        fontSize="10px"
+        lineHeight="16px"
+        fontWeight="normal"
+        letterSpacing="0.3px"
+        textTransform="uppercase"
+        color="#656e7b"
```
guilherme-n: do you think we could get this color from the chakra theme?

app/components/PageHeader.tsx
```diff
@@ -109,27 +207,23 @@ function PageHeader() {
...
-            {usedPrompts}/
-            {totalPrompts > 5000 ? (
-              <Text as="span" fontSize="xl" verticalAlign="bottom">
-                ∞
-              </Text>
-            ) : (
-              totalPrompts
-            )}{" "}
-            daily prompts
+            {usedPrompts} / {totalPrompts > 5000 ? "∞" : totalPrompts} prompts
```
guilherme-n: We used to call it "daily prompts". Just checking if that's intentional, since the limits still apply

guilherme-n (approved):

app/components/PageHeader.tsx
01painadam: I was following the design, but yeah I feel like we should be explicit

PR #511, opened 2026-06-02
add a Stop button to cancel the current chat, firing the AbortController

batpad:
Adds a "Stop" button which then fires the abort controller to cancel the streaming chat request and immediately end the chat.

I didn't really know what I was doing here, and wanted more to test that this works. Am still a bit confused on whether it's properly doing the cancel on the backend, might be nice to get more eyes.

Am pushing this as a draft PR, but I think ideally it would be nice if @kamicut or @guilherme-n were able to look and re-implement / implement nicer? 

This should work to test though - adds a little Stop button to the chat, and clicking on it should fire the Abort Controller to cancel the request.

cc @01painadam


01painadam:
Thanks @batpad, this is great! Gonna pass this by design and update the ticket

Soljeri:
UI feedback: 
One additional approach could be to follow a pattern similar to chatGPT/Claude ’s message actions by surfacing contextual actions directly on the message bubble through icon only controls (e.g., a pencil icon for Edit and a refresh/retry icon for Regenerate). 

Mainly this would improve discoverability and reduce ambiguity—we could pair the icons with lightweight tooltips on hover/focus.

01painadam:
> UI feedback: One additional approach could be to follow a pattern similar to chatGPT/Claude ’s message actions by surfacing contextual actions directly on the message bubble through icon only controls (e.g., a pencil icon for Edit and a refresh/retry icon for Regenerate).
> 
> Mainly this would improve discoverability and reduce ambiguity—we could pair the icons with lightweight tooltips on hover/focus.

@kamicut @batpad this is not blocking - gonna add a ticket

PR #513, opened 2026-06-02
PZB-904: Add Area Card to chat on AOI selection

guilherme-n:
## Pull request type
- [x] Feature

When the AI selects an area of interest (AOI), no visual confirmation appears in the chat thread. The user has no immediate feedback about which area was selected or what type it is, and must look at the map to understand what happened.

## What is the new behavior?
A new Area Card appears in the chat thread immediately after the AI selects an AOI, showing a Mapbox static map thumbnail of the area, its name, source type (e.g. "Administrative Areas"), and a crosshair icon button to re-center the map on it
The DatasetCard component was refactored to compose the existing InfoCard primitive, removing ~90 lines of duplicated layout code; AreaCard and AnalysisCard also share the same InfoCard base — all three card variants are now consistent
The unused size prop was removed from DatasetCard

- [x] No

## Demo
Screenshot 2026-06-02 at 17 16 36

### DIFF
--- a/app/components/AreaCard.tsx
+++ b/app/components/AreaCard.tsx
@@ -1,0 +1,242 @@
+"use client";
+import { Box, Image } from "@chakra-ui/react";
+import { CrosshairIcon, PolygonIcon } from "@phosphor-icons/react";
+import { AOISelection } from "@/app/types/chat";
+import { Tooltip } from "./ui/tooltip";
+import { InfoCard } from "./InfoCard";
+import useMapStore from "@/app/store/mapStore";
+import { unionAoiBboxes } from "@/app/utils/bboxUtils";
+import { FeatureCollection, Feature, Geometry } from "geojson";
+import { useState } from "react";
+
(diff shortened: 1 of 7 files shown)


app/components/AreaCard.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  // Fall back to bounds-only (no overlay)
+  return `${baseUrl}/${boundsPath}${qs}`;
+}
+
+function getAreaTypeLabel(aoiSelection: AOISelection): string {
+  const aoi = aoiSelection.aois[0];
+  if (!aoi) return "";
+  if (aoi.subtype === "custom-area") return "Custom Polygon";
+  switch (aoi.source) {
```
willian-viana, line 144: nit: I'd recommend separating blocks of code such as if, switch, and constants with blank lines.

willian-viana (approved):
:shipit:

app/components/AreaCard.tsx
guilherme-n: let's try to have some team agreement and put the style standards on prettier to make it easier =)

PR #514, opened 2026-06-03
PZB-835: Hotfix/loosen insight card matching criteria

01painadam:
## Problem

Insight cards stopped rendering inline in the chat thread. Cards only appeared in the map-overlay
InsightWorkspace, not in the conversation where users expect them.

Root cause seems to be FE/BE contract drift which changed insight rendering to a two-part handshake:

1. The generate_insights tool message carries charts_data → generateInsightsTool populates
insightStore (insights[] + pendingBatch).
2. The assistant reply text must then contain `[Chart N]` or `[Chart {uuid}]` markers, which the FE detects and uses to inject cards into chat

Staging currently emits the chart data but no `[Chart ...]` markers in the reply.

## Fix

A frontend bridge that uses what staging already emits (the chart data), with no backend
dependency:

- New `app/lib/insight-chat-messages.ts` extracts the _assistant-message → chat-messages_ logic into a
pure, tested helper with two paths:
  - Marker path (unchanged): when [Chart N] markers are present, cards are placed positionally.
Preserved exactly, so when the backend marker change deploys it takes over automatically — no
further FE change needed. (The regex matches both UUID and integer-index marker forms.)
  - Fallback path (new): when there are no markers but a generate_insights batch is pending,
render the cards immediately before the narrative text, so the text reads as a conclusion to the
chart card above it.
- `app/store/chatStore.ts` now uses a single call into the helper and `pendingBatch` is now always consumed in the text branch.

## Testing

- app/lib/__tests__/insight-chat-messages.test.ts — 4 new tests covering both paths
(cards-before-text fallback, text-only when no cards, positional marker injection,
unmatched-marker skip).
- pnpm test (27/27), pnpm typecheck, pnpm lint, pnpm format:check all pass.

Should observe the following:
Screenshot 2026-06-03 at 09 44 17

### DIFF
--- a/app/lib/__tests__/insight-chat-messages.test.ts
+++ b/app/lib/__tests__/insight-chat-messages.test.ts
@@ -1,0 +1,147 @@
+import { describe, it, expect } from "vitest";
+import { buildInsightChatMessages } from "@/app/lib/insight-chat-messages";
+import { InsightWidget } from "@/app/types/chat";
+
+const TS = "2026-06-02T00:00:00.000Z";
+const TRACE = "trace-123";
+
+function widget(id: string, title: string): InsightWidget {
+  return {
+    id,
+    type: "bar",
+    title,
+    description: "",
+    data: [],
+    xAxis: "x",
+    yAxis: "y",
+  };
+}
+
+describe("buildInsightChatMessages", () => {
+  describe("fallback path (no [Chart N] markers — current staging)", () => {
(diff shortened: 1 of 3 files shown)


gtempus (commented):
Definitely +1 for moving that code out of the store, @01painadam 💯 💪

guilherme-n (approved):

PR #515, opened 2026-06-03
Add default start and end years

01painadam:
Adds start AND end years for manually selected temporal data.

- select TCL or TCL fires from dataset catalogue
- check that tiles return 2001-2025


willian-viana (approved):
:shipit:

PR #516, opened 2026-06-04
[PZB-900] Support Vector Tile endpoints for context layers

kamicut:
Context layers have up til now been only raster. This PR adds vector tile (MVT/.pbf) rendering when the backend provides a `source_layer` alongside `tile_url`. 

**How it works**

When `context_layers[]` contains a `source_layer` field (or an explicit type: "vector"), the pipeline now:

1. Skips the primary-forest PNG protocol wrapper on the URL we had added for black pixels (This is tech debt @01painadam we should resolve, I've added tests here to ensure that the patch is not used on the vector tile layers)
2. Creates the sub-layer as type: "vector" in the layer store with a `vectorStyle` descriptor resolved from `CONTEXT_LAYER_METADATA`
3. Renders it through a new VectorDataLayers component that uses `vectorStyle`
4. For IFL specifically, only year == 2000 features render filled #5C8C50; all other years are transparent

The vector data layer renderer is generic in that it could be used for future vector tile data that are not contextual layers. If a layer has a "parentId", then it gets nested as context within the parent, similar to the approach for raster layers.

I've also added mock data to the "Debug Toast Panel", although we should probably rename that panel @01painadam. It also gives us a model of how to call tools directly if we wanted to test it.


app/components/map/layers/AoiVectorTileLayers.tsx
```diff
@@ -13,13 +13,16 @@ interface VectorTileLayersProps {
...
   const allLayers = useMapStore((s) => s.layers);
   const vectorLayers = useMemo(
     () =>
       allLayers.filter(
         (l): l is ManagedLayer & { tileUrl: string; sourceLayer: string } =>
-          l.type === "vector" && !!l.tileUrl && !!l.sourceLayer
+          l.type === "vector" &&
+          !!l.tileUrl &&
+          !!l.sourceLayer &&
+          !l.vectorStyle
```
willian-viana: I believe there's a typo here. Shouldn't it be `!!l.vectorStyle`?

willian-viana (approved):
:shipit:

kamicut:
@willian-viana after your review I rewrote that line - it was not very clear how we were making distinctions between AOI vector tiles and dataset vector tiles. I added a couple of utility functions for it.

PR #520, opened 2026-06-05
PZB-948 - Whats new header UI

01painadam:
### Summary
Replaces the blue header style background with white + lime-green top border, and adds a **What's new** button with notification indicator.

Moves WhatsNewModal from ChatMessages (only mounted when messages exist) to the chat layout, so it is always available regardless of conversation state. Auto-open on first visit is removed and the panel is now purely opt-in via the header button.

image

### DIFF
--- a/app/components/PageHeader.tsx
+++ b/app/components/PageHeader.tsx
@@ -38,6 +38,7 @@ import { useLogout } from "@/app/hooks/useLogout";
 
 const isPrototype = process.env.NEXT_PUBLIC_PROTOTYPE_MODE === "true";
 const DISCLAIMER_STORAGE_KEY = "gnw_disclaimer_dismissed_v2";
+const WHATS_NEW_STORAGE_KEY = "whats-new-v3-dismissed";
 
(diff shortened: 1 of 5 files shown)


app/components/PageHeader.tsx
```diff
@@ -52,8 +54,8 @@ function PageHeader() {
     ? currentThread.name
     : "New Conversation";
 
-  const inverseColor = isPrototype ? "#1f2937" : "fg.inverted";
-  const inverseHoverBg = isPrototype ? "#6b7280" : "primary.fg";
+  const inverseColor = isPrototype ? "#1f2937" : "#3A4048";
+  const inverseHoverBg = isPrototype ? "#6b7280" : "#F0F1F2";
```
guilherme-n: Would you know if we already have these colors in the theme?

guilherme-n (approved):

app/components/PageHeader.tsx
01painadam: Maybe, I'll take a look

PR #528, opened 2026-06-08
PZB-955: enable GADM layer as active by default on landing

guilherme-n:
## Pull request type

- [x] Feature

## What is the new behavior?
Landing on the app with no prior session state shows GADM active and clickable, making areas immediately selectable without any setup.

- [ ] Yes
- [x] No

## Demo
Screenshot 2026-06-08 at 22 23 07

### DIFF
--- a/app/components/MapAreaControls.tsx
+++ b/app/components/MapAreaControls.tsx
@@ -70,6 +70,7 @@ function MapAreaControls({
   setBasemapTiles,
 }: MapAreaControlsProps) {
   const {
+    selectAreaLayer,
     setSelectAreaLayer,
     isDrawingMode,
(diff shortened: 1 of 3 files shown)


01painadam:
Looks great! Two small things:
- Can we make it so you can only select one AoI at a time (selecting a second, deselects the first). This mirrors GFW and avoids issues analysing many AOIs at once
- Since (GADM) admin areas are now selected by default, can we disable it in the area tools drop down until another AOI layer is selected? Maybe just grey out the text? See below 

Screenshot 2026-06-09 at 8 45 38 AM

01painadam (approved):

PR #532, opened 2026-06-11
PZB-946: Replace cookies notice modal with updated design

guilherme-n:
## Pull request type

- [x] Feature

## What is the new behavior?
Cookies notice modal now uses updated design.

- [x] No

## Demo
Screenshot 2026-06-10 at 21 25 49
Screenshot 2026-06-10 at 21 25 07


01painadam (approved):
Looks great! One minor comment, I wonder if we should position the modal (vertically) on top of the legend?

Maybe worth a quick sync with Sol

Otherwise, LGTM!

PR #536, opened 2026-06-12
feat(chat): render WRI Insights citations as hover-card chips

yellowcap:
## Summary

Assistant replies now cite WRI Insights articles with compact numbered markdown links like `[1](https://www.wri.org/insights/...)`. This PR replaces each marker with an inline citation chip (WRI favicon + number) that shows a rich article card on hover.

- **`BlogCitation` component** — superscript chip rendered in place of `[N](url)` markers; on hover shows a card with the article's image, title, abstract and last-modified date, with skeleton loading and a graceful fallback for unrecognized URLs. Both the chip and the card open the article in a new tab.
- **`useBlogCitations` hook** — extracts all cited URLs from a message and resolves them in a single batched request to the new `GET /api/blogs/metadata` backend endpoint (React Query, cached per URL set, Zod-validated).
- **`MessageBubble`** — custom `react-markdown` link renderer for assistant messages: citation links become chips, all other links render unchanged (existing link styles scoped with `:not([data-citation])`).
- **`app/lib/blog-citations.ts`** — pure helpers for detecting citation links and extracting cited URLs, with unit tests.
- **`public/wri-favicon.ico`** — WRI favicon served locally rather than hotlinked.

Requires the `project-zeno` backend with the `/api/blogs/metadata` router (wri/project-zeno blog search feature).
https://github.com/wri/project-zeno/pull/707

Screenshot of this draft feature, pending actual design
image


01painadam (approved):
Functionally, works as expected. Good to go to staging pending a proper code review

app/lib/blog-citations.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/** True when a markdown link is a numbered Insights citation marker. */
+export function isBlogCitation(href: string, label: string): boolean {
+  return CITATION_LABEL_RE.test(label.trim()) && INSIGHTS_URL_RE.test(href);
+}
+
+/** Unique Insights URLs cited (as `[N](url)`) in a markdown string. */
+export function extractCitationUrls(markdown: string): string[] {
+  const urls = new Set<string>();
+  for (const match of markdown.matchAll(CITATION_LINK_RE)) {
```
kamicut, line 44: this regex defers from `INSIGHTS_URL_RE`, and might break if insight URLs are at nested paths (say `/insights/foo/bar` instead of `/insights/foo`). This would make `isBlogCitation` and `extractCitationUrls` disagree. This is low severity if we assume insights are always at `/insights/foo` (single segment URL and no extra nesting).

kamicut (approved):
I'm flagging a low-severity bug that might happen if assumptions about citation URLs change @yellowcap

app/lib/blog-citations.ts
01painadam: Nice, I'll add a todo to make this source dependent

PR #551, opened 2026-06-19
Update map aoi tooltip

01painadam:
- Updates tooltip styles on-hover
- add loading state '...' (unused in this layer but preparing for future layers)
- BONUS - disable AOI layer when drawing


willian-viana (approved):
:shipit:

PR #554, opened 2026-06-22
Rebrand headers + What's new

01painadam:
Rebrand changes GNW --> GNW Horizon across the app (excluding landing page copy)


app/components/WelcomeModal.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO: Remove this component when the app
```
guilherme-n: seems like this phrase is incomplete

guilherme-n (approved):

PR #557, opened 2026-06-23
User can explore Dataset catalogue

guilherme-n:
## Pull request type

- [x] Feature

On desktop, users browse and add dataset layers through the chat input’s Data catalog button, which opens a centered modal (ContextMenu) with side tabs for Data catalog and Area tools. Layers can also be added by asking the assistant. There is no dedicated catalog panel in the map/exploration layout — dataset picking lives inside that modal alongside area selection. Map controls (zoom, basemap) are positioned relative to the chat panel only, and the Area tools flow shares the same modal navigation as the dataset catalogue.

## What is the new behavior?
Users can browse and add dataset layers from a dedicated Data catalog panel (400px wide) in the desktop exploration layout, without going through the assistant.

- Opening the panel: The existing Data catalog button in the chat tool row toggles the panel on desktop. On mobile, the button still opens the legacy modal.
- Layout: In compact chat mode, the catalog sits on the far left and pushes the chat panel right so they don’t overlap. In full-size chat mode, the catalog docks immediately beside the chat panel with a gray divider between them. Map controls shift right when the catalog is open.
- Catalog contents: Category filter chips (All datasets, In this conversation, Land use, etc.) narrow the list. Each dataset is shown as a 376×115px card with thumbnail, DATA label, title, metadata (cadence · extent · source), info affordance, and a Show on map toggle.
- Map integration: Toggling Show on map adds/removes layers via the existing contextStore → mapStore wiring. Active cards expose inline visibility and opacity controls. Layers still appear in the existing legend panel with full functionality.
- Area tools: The Area tools modal no longer shows the Data catalog tab — area selection is isolated from the new side panel.
- Polish: The card list scrolls without visible scrollbars, and horizontal overflow in the panel is suppressed.

- [x] No

## Demo
Screenshot 2026-06-23 at 14 59 31

---

Screenshot 2026-06-23 at 14 59 42

### DIFF
--- a/app/ChatPanelCompact.tsx
+++ b/app/ChatPanelCompact.tsx
@@ -9,8 +9,10 @@ import ChatPanelHeader from "./ChatPanelHeader";
 import ChatPanelDisclaimer from "./ChatPanelDisclaimer";
 import PromptQuotaNotice from "./PromptQuotaNotice";
 import { chatPanelCardStyle } from "./chatPanelShared";
+import { getCompactChatLeftPx } from "./explorationLayout";
 import { usePromptQuota } from "./hooks/usePromptQuota";
 import useChatStore from "./store/chatStore";
+import useSidebarStore from "./store/sidebarStore";
 import { useState, useEffect, useRef, useCallback } from "react";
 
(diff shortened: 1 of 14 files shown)


willian-viana (approved):
LGTM :shipit:

PR #558, opened 2026-06-23
Insight workspace loading

01painadam:
I noticed that the Insight loading state was being triggered by any prompt, even ones that don't need generate an insight e.g. `What datasets related to fires do you have?`

I introduced an `isGeneratingInsight` flag that is true only between the agent message contains a `generate_insights` tool call, which signals that an insight is loading. The agent emits its tool-call decision as a separate stream event before the chart data streams back, so this is the earliest correct moment to surface the loading state.

### Changes

- `types/chat.ts` adds `tool_calls?: string[]` to StreamMessage.
- `parse-stream-message.ts` now surfaces the agent's announced tool_calls names: attached to the text message when the turn includes narration, or emitted as a lightweight other/tool_calls message for a pure tool-call turn (previously dropped entirely).
- `chatStore.ts` adds isGeneratingInsight state + setGeneratingInsight action. Set true when
generate_insights is announced; cleared in the same task that adds the chart (skeleton → chart swap in a single render, no empty flash), on a generate_insights error, at send-start, and in both sendMessage/fetchThread finally blocks as a safety net.
- `InsightWorkspace.tsx` gates both the skeleton and the "INSIGHTS LOADING" header on isGeneratingInsight instead of isLoading.

### New Behaviour

- The skeleton/spinner appear only once generate_insights is announced.
- The workspace stays empty (null) for non-insight prompts.
- The "INSIGHTS LOADING" header no longer fires on follow-up non-insight prompts when an insight card already exists.
- Thread replay unaffected: the backend filters empty-content AI messages on replay, but replay doesn't need the skeleton charts are already populated when the generate_insights result lands.

### DIFF
--- a/app/components/InsightWorkspace.tsx
+++ b/app/components/InsightWorkspace.tsx
@@ -101,7 +101,12 @@ function WorkspaceSkeleton() {
 
 export default function InsightWorkspace() {
   const { insights } = useInsightStore();
-  const isLoading = useChatStore((state) => state.isLoading);
+  // Drive the loading affordances off the insight-specific flag, not the
+  // request-wide isLoading: the skeleton/spinner should appear only while an
+  // insight is actually being generated, not for every prompt.
+  const isGeneratingInsight = useChatStore(
+    (state) => state.isGeneratingInsight
+  );
   const [currentIndex, setCurrentIndex] = useState(0);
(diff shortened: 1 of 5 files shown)


willian-viana (approved):
LGTM :shipit:

PR #562, opened 2026-06-24
PZB-1032: Refactor/remove context store

kamicut:
# Remove `contextStore`
> **Draft.** Structural decoupling + a behavior-preserving `ui_context` port. The
> `ui_context` contract redesign is deferred to the workspace-awareness spike.

## Why

The context store had become a duplicate selection store: `mapStore.layers` already
owns the real state (geometry, AOI, dataset, params, visibility). An area selection
lived in three places kept in sync by fragile string-matching, plus an `isInContext`
flag that just mirrored "is this layer visible."

This PR removes the context store and re-homes some of its responsibilities. Input pills 
become a view of visible layers + date. This will change in future designs where we remove
the pills. The wire contract to the backend does **not** change.

## Tasks

- [x] **Datasets off context** — datasets are now driven entirely by map layers
      (add/remove via layer menu, dataset card, nudge, legend stay in sync; analysis
      CTA + export read from layers).
- [x] **Areas off context** — visible area layers are the scope; delete the parallel
      area state and the blue/gray highlight logic.
- [x] **Date → `chatStore`** — the one concern with no map counterpart.
- [x] **`ui_context` port** — re-home the derivation off `contextStore` (byte-identical),
      add a per-message scope snapshot, and rehydrate scope on thread load.
- [x] **Delete `contextStore`** + update docs.

### Deferred to the workspace-awareness spike (not this PR)
Richer `ui_context` contract: evolving `ui_context` into a structured workspace representation.

## ⚠️ Flags for others working in this area

- **Visible layer = in scope.** There is no `isInContext`/selected concept anymore. If you
  add a layer to the map, it's in the query scope by definition.
- **Backend has no deselect.** Removing the only area layer does not tell the backend to drop
  the AOI; it keeps the last one
- **`buildDatasetLayers` is the single entry point** for putting a dataset on the map (main +
  optional context sub-layer). Don't reintroduce a parallel "selection" store.


app/store/chat-tools/generateInsights.ts
```diff
@@ -35,6 +36,13 @@ export function generateInsightsTool(
       // analysis. Prefer it over contextStore, which is updated asynchronously
       // by pickAoiTool/etc. and may not yet reflect this turn's choices.
       const context = useContextStore.getState().context;
```
willian-viana: Are you not removing `useContextStore` fully?

app/components/ContextMenu.tsx
```diff
@@ -80,29 +81,27 @@ function LayerCardList({
...
-        contextType: "layer",
-        ...getLayerContextFromDatasetCard(card),
-        isAiContext: false,
-      });
+    const isSelected = layers.some((l) => l.datasetId === card.dataset_id);
+    // Single-dataset selection: clear any existing dataset layers, then add the
+    // clicked one — unless it was already selected, in which case this is a
+    // toggle-off. The visible layer IS the scope (no separate context item).
+    layers
+      .filter((l) => typeof l.datasetId === "number")
```
willian-viana: This logic is repeatedly being used across this refactor. Would it be a good idea to have an util for it?

app/components/AreasPanel.tsx
```diff
@@ -243,64 +246,51 @@ function AreasPanelHeader({ onClose }: { onClose: () => void }) {
 }
 
 /** Type label string ("Administrative areas" etc.) derived from an area context item. */
-function getAreaTypeLabel(item: ContextItem): string {
-  const source =
-    item.aoiData?.source ?? item.aoiSelection?.aois[0]?.source ?? "";
+function getAreaTypeLabel(layer: Layer): string {
```
willian-viana: I believe reading `layer.aoiSelection` directly would mislabel every hand-clicked boundary (GADM) as "User upload areas", since manually-selected areas don't have `aoiSelection`. 

```suggestion
function getAreaTypeLabel(aoiSelection: AOISelection | null): string {
  const source = aoiSelection?.aois[0]?.source ?? "custom";
```

willian-viana (changes requested):
This looks great! Thanks for this large refactor @kamicut 🚀 
I suggested two small (but perhaps important) changes.

willian-viana (approved):
:shipit:

PR #565, opened 2026-06-26
PZB-774: explore AOI tools button areas

guilherme-n:
## Pull request type

- [x] Feature

The Areas button in the chat input opens a modal (ContextMenu) to browse and add areas. Upload, select-on-map, and draw controls live as floating buttons on the map overlay (MapAreaControls), alongside zoom and basemap controls. Selection-mode banners and draw validation errors also render inside that map overlay, which can be clipped or misaligned when the Data Catalog panel is open. The exploration layout only accounts for the catalog column when shifting chat and map chrome, there is no dedicated areas panel, and only one slide-out column slot exists (catalog only).

## What is the new behavior?
The Areas button toggles a new Areas slide-out panel that shares the same column slot as the Data Catalog (opening one closes the other). The panel lists areas in two tabs — In this conversation (from chat context, with show/hide and remove) and Monitored areas (saved custom areas, with add-to-conversation and map visibility). Upload, select-on-map, and draw tools move into the panel header; the map overlay keeps only zoom, basemap, and mobile tools. Selection-mode and draw-validation feedback move to a dedicated MapAreaFeedback layer so they stay visible above the catalog/areas column. Chat, map controls, and feedback reposition consistently when either panel is open, matching the existing catalog layout behavior. On mobile, area picking still uses the modal — only desktop gets the new panel.

- [x] No

## Demo
Screenshot 2026-06-25 at 22 33 40


01painadam (approved):
LGTM!

PR #574, opened 2026-07-06
feat(chat): pin active user prompt to top of message list

01painadam:
## Summary
- Adds a sticky `PinnedPrompt` overlay that keeps the in-flight user prompt visible while the response streams in, with a jump-back affordance to scroll to the original message.
- New `usePinnedPrompt` hook tracks the last user message's position via `IntersectionObserver` and derives pinned state.
- Pure helpers in `app/utils/pinnedPrompt.ts` with unit tests.
- Moves scroller top padding into `ChatMessages` (sticky offsets would otherwise include the scroll container's padding).
- `MessageBubble` accepts a ref and drives first-child top margin off the `isFirst` prop so the overlay doesn't disturb spacing.

Originally landed directly on `develop` as c7a3c9d without a PR; that commit has been reverted on `develop` and reapplied here via cherry-pick for review.

image

## Test plan
- [ ] `pnpm test` passes (includes `app/utils/__tests__/pinnedPrompt.test.ts`)
- [ ] `pnpm typecheck` / `pnpm lint` / `pnpm format:check` pass
- [ ] Manually verify in the full-size and compact chat panels: send a message, confirm it pins to the top while the response streams, and the jump-back affordance scrolls to the original message

### DIFF
--- a/app/ChatPanelCompact.tsx
+++ b/app/ChatPanelCompact.tsx
@@ -133,15 +133,16 @@ function ChatPanelCompact({ onToggleSize }: ChatPanelCompactProps) {
                 transition={{ duration: 0.22, ease: "easeInOut" }}
                 style={{ overflow: "hidden" }}
               >
+                {/* Top padding is passed to ChatMessages instead of set on
+                    this scroller — see ChatMessagesProps.pt. */}
                 <Box
                   ref={messagesRef}
                   overflowY="auto"
                   px={4}
-                  pt={4}
                   pb={4}
                   maxH={messagesMaxH}
                 >
-                  <ChatMessages />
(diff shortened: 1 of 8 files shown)


guilherme-n:
I haven't seen the figma mocks (if they do exist for this one), but I have the impression we could use some gap between the blue container and the text below it 
Screenshot 2026-07-08 at 13 34 03

guilherme-n (approved):

01painadam:
> I haven't seen the figma mocks (if they do exist for this one), but I have the impression we could use some gap between the blue container and the text below it Screenshot 2026-07-08 at 13 34 03

Yeah, good point - I'm gonna merge this into develop and ask Sol for a design review.

PR #578, opened 2026-07-08
feat(chat): add voice-to-text (dictation) input flagged begind ff=voice

01painadam:
## What & why

Adds voice-to-text dictation to the chat input so users can speak their query instead of typing it. **No backend change** — the transcript flows into the same `sendMessage(text)` path as typing, so `/api/chat` sees plain text either way.

## Approach

Uses the **browser-native Web Speech API** (`SpeechRecognition` / `webkitSpeechRecognition`) — the light-touch option: no new dependency, no API key, no cost, and no audio leaves the browser's own speech service. This keeps the change entirely frontend-side.

## What changed

| File | What |
|---|---|
| `app/types/speech-recognition.d.ts` | Ambient TS types for the Web Speech API (not in the default DOM lib) |
| `app/hooks/useSpeechInput.ts` | `useSpeechInput` hook wrapping `SpeechRecognition`; returns `null` when unsupported |
| `app/components/ChatInput.tsx` | Mic button beside the send button, feeding the existing `inputValue` |

## Behaviour

- **Mic button** sits left of the send button. Idle = ghost mic; recording = solid red mic-slash. Tap to start/stop.
- **Dictation appends** to any already-typed text (input is snapshotted on start) and updates live as you speak.
- **Stops on send** — `submitPrompt` calls `speech?.stop()`.
- **Degrades cleanly** — the button only renders where the API exists (Chrome/Edge/Safari). Firefox and SSR get no button, and support is resolved in an effect to avoid a hydration mismatch.

## Verification

- `pnpm typecheck` — clean
- `pnpm build` — passes
- Prettier — clean
- Pre-push hook (format:check + typecheck) — passed

# Voice-to-Text (Dictation) — Languages & Browser Coverage

Speak instead of type. The mic button in the chat prompt streams speech to a
transcript that lands in the input box.

> **In one breath:** dictation runs entirely in the browser via the native Web
> Speech API. No audio and no transcript ever reach the GNW backend — so which
> languages work, and how well, is decided by the _browser_, not by us.

This page documents the language catalogue we offer and how coverage differs
across browsers. It is the source of truth for the two questions people ask
most: _"which languages can I dictate in?"_ and _"why is the mic button
missing / not working in my browser?"_

## How it works (why coverage is browser-dependent)

The feature is a thin wrapper around the browser-native
[`SpeechRecognition`](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition)
interface (`window.SpeechRecognition ?? window.webkitSpeechRecognition`).

- **No backend involvement.** Transcription happens client-side. In Chromium
  browsers the audio is sent to Google's speech service; in Safari it uses
  Apple's on-device / dictation engine. Either way it never touches our API.
- **The API cannot detect the spoken language.** We must hand it a BCP-47 tag
  up front (see [Language resolution](#language-resolution--defaults)).
- **If the API is absent, the mic button is hidden.** `useSpeechInput` returns
  `null` when neither constructor exists, so the affordance simply doesn't
  render (e.g. in Firefox).

Because the actual recognition engine is the browser's, **the set of languages
that truly work — and the transcription quality — is determined by the browser
and the user's device, not by our catalogue.** Our catalogue is the _menu_ we
offer; the browser decides what it can cook.

Source of truth in the code:

| Concern            | File                                                      |
| ------------------ | --------------------------------------------------------- |
| Language catalogue | `app/utils/speechLang.ts` (`VOICE_LANGUAGE_FAMILIES`)     |
| Support detection  | `app/hooks/useSpeechInput.ts` (`getCtor`, `getSupported`) |
| Picker UI          | `app/components/voice/ShortlistLanguagePicker.tsx`        |

---

## Browser coverage

Support is gated on the presence of the `SpeechRecognition` /
`webkitSpeechRecognition` constructor. This is the primary "is the mic button
even there?" gate.

| Browser                                    | Web Speech API                 | Behaviour in GNW                                                                                                                                      |
| ------------------------------------------ | ------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Chrome** (desktop & Android) 25+         | ✅ `webkitSpeechRecognition`   | Full catalogue. Audio routed to Google's speech service.                                                                                              |
| **Edge** 87+                               | ✅ `webkitSpeechRecognition`   | Full catalogue (Chromium, same Google backend).                                                                                                       |
| **Opera / other Chromium**                 | ✅ `webkitSpeechRecognition`   | Full catalogue (Chromium, same Google backend).                                                                                                       |
| **Samsung Internet** 4+                    | ✅ `webkitSpeechRecognition`   | Full catalogue.                                                                                                                                       |
| **Safari** (macOS 14.1+, iOS/iPadOS 14.5+) | ✅ `webkitSpeechRecognition`   | Mic button shows. Uses Apple's dictation engine — language coverage is a **device-dependent subset** (see below).                                     |
| **Brave**                                  | ⚠️ present but usually blocked | Constructor exists, so the button shows, but Brave disables Google's speech endpoint by default, so recognition typically fails with a service error. |
| **Firefox**                                | ❌ off by default              | `SpeechRecognition` sits behind the `dom.webspeech.recognition.enable` flag and ships no engine. **Mic button is hidden.**                            |
| **Server-side render (SSR)**               | ❌ n/a                         | Support resolves to `false` on the server and first client render to avoid a hydration mismatch, then re-resolves in the browser.                     |

### Per-language coverage, by engine

The catalogue below lists every language we _offer_. Whether a given language
actually transcribes depends on the engine behind the browser:

- **Chromium browsers (Chrome, Edge, Opera, Samsung Internet)** — the **entire
  catalogue** is supported. Every code in `speechLang.ts` was chosen because
  it's a locale Chrome's `SpeechRecognition` accepts.
- **Safari (WebKit)** — recognition is delegated to Apple's dictation engine,
  so a language works only if its **dictation locale is available on that
  device**. Apple supports 50+ dictation locales, which covers **all eight
  shortlist ("common") languages and the large majority of the extended list**.
  Regional variants Apple does not distinguish (e.g. `es-419`, and `en-GB`
  vs `en-US`) resolve to the closest base locale rather than failing. Treat
  Safari as _"all common languages + most extended languages, subject to the
  device's installed dictation languages."_
- **Firefox** — no languages; the feature is hidden entirely.

> **Rule of thumb:** if a listed language ever fails to transcribe, it's a
> Safari/Firefox engine limitation, not a bug in our catalogue. In Chrome the
> full list is expected to work.

---

## Language catalogue

**35 language families → 39 BCP-47 locale codes.** The 8 families marked ●
form the "Common" shortlist shown at the top of the picker before the user
searches or expands the full list; the rest appear under the full list.

### Shortlist ("Common")

| ●   | Language   | Native           | BCP-47 code(s)                              |
| --- | ---------- | ---------------- | ------------------------------------------- |
| ●   | English    | English          | `en-US` (US), `en-GB` (UK)                  |
| ●   | Spanish    | Español          | `es-ES` (Spain), `es-419` (Latin America)   |
| ●   | Portuguese | Português        | `pt-BR` (Brasil), `pt-PT` (Portugal)        |
| ●   | Chinese    | 中文             | `zh-CN` (Simplified), `zh-TW` (Traditional) |
| ●   | French     | Français         | `fr-FR`                                     |
| ●   | Arabic     | العربية          | `ar-SA`                                     |
| ●   | Hindi      | हिन्दी           | `hi-IN`                                     |
| ●   | Indonesian | Bahasa Indonesia | `id-ID`                                     |

### Extended list

| Language   | Native        | BCP-47 code |
| ---------- | ------------- | ----------- |
| Bengali    | বাংলা         | `bn-IN`     |
| Czech      | Čeština       | `cs-CZ`     |
| Danish     | Dansk         | `da-DK`     |
| Dutch      | Nederlands    | `nl-NL`     |
| Filipino   | Filipino      | `fil-PH`    |
| Finnish    | Suomi         | `fi-FI`     |
| German     | Deutsch       | `de-DE`     |
| Greek      | Ελληνικά      | `el-GR`     |
| Gujarati   | ગુજરાતી       | `gu-IN`     |
| Hungarian  | Magyar        | `hu-HU`     |
| Italian    | Italiano      | `it-IT`     |
| Japanese   | 日本語        | `ja-JP`     |
| Kiswahili  | Kiswahili     | `sw-KE`     |
| Korean     | 한국어        | `ko-KR`     |
| Malay      | Bahasa Melayu | `ms-MY`     |
| Norwegian  | Norsk         | `nb-NO`     |
| Polish     | Polski        | `pl-PL`     |
| Romanian   | Română        | `ro-RO`     |
| Russian    | Русский       | `ru-RU`     |
| Swedish    | Svenska       | `sv-SE`     |
| Tamil      | தமிழ்         | `ta-IN`     |
| Telugu     | తెలుగు        | `te-IN`     |
| Thai       | ภาษาไทย       | `th-TH`     |
| Turkish    | Türkçe        | `tr-TR`     |
| Ukrainian  | Українська    | `uk-UA`     |
| Urdu       | اردو          | `ur-PK`     |
| Vietnamese | Tiếng Việt    | `vi-VN`     |

_All codes above are supported in Chromium browsers. Safari support tracks the
device's installed Apple dictation locales (see
[Per-language coverage](#per-language-coverage-by-engine)); Firefox supports
none._

---

## Language resolution & defaults

The Web Speech API can't guess the spoken language, so `resolveSpeechLang`
picks a starting tag in priority order:

1. The user's onboarding language preference (`preferredLanguageCode`)
2. The browser's preferred language (`navigator.language`)
3. `en-US`

A bare ISO-639-1 code (e.g. `pt`) is region-qualified for better accent
handling via a small map — `en→en-US`, `fr→fr-FR`, `es→es-ES`, `pt→pt-BR`,
`id→id-ID`. An already-qualified tag (e.g. `en-GB`) is trusted as-is.

The user can override the language for the current session from the language
menu in the listening footer (`VoiceLanguageMenu` → `ShortlistLanguagePicker`).
Switching language mid-session tears down and restarts recognition in place,
carrying the transcript captured so far.

---

## Privacy note

Because transcription is browser-native:

- **Chromium:** audio is streamed to Google's speech-recognition service.
- **Safari:** audio is handled by Apple's on-device / dictation engine.
- **GNW backend:** receives neither the audio nor the interim transcript. Only
  the final text the user chooses to send (as a normal chat message) leaves the
  browser.

---

_Last updated: 2026-07-09. Catalogue counts and support logic verified against
`app/utils/speechLang.ts` and `app/hooks/useSpeechInput.ts`._

### DIFF
--- a/app/components/ChatInput.tsx
+++ b/app/components/ChatInput.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { useState } from "react";
+import { useRef, useState } from "react";
 import {
   Button,
   Flex,
@@ -10,7 +10,12 @@ import {
   Text,
   Portal,
 } from "@chakra-ui/react";
-import { ArrowBendRightUpIcon, StopIcon } from "@phosphor-icons/react";
+import {
+  ArrowBendRightUpIcon,
+  MicrophoneIcon,
+  MicrophoneSlashIcon,
+  StopIcon,
+} from "@phosphor-icons/react";
 import { format } from "date-fns";
 import useChatStore from "@/app/store/chatStore";
 import ContextButton, { ChatContextType } from "./ContextButton";
@@ -19,6 +24,7 @@ import ContextMenu from "./ContextMenu";
 import useMapStore from "../store/mapStore";
 import { isAreaLayer } from "../store/layerManagerSlice";
 import useSidebarStore from "../store/sidebarStore";
+import useSpeechInput from "../hooks/useSpeechInput";
 import { useRouter } from "next/navigation";
 
 export default function ChatInput({
@@ -70,6 +76,20 @@ export default function ChatInput({
   const excludedLayerIds = useChatStore((s) => s.excludedContextLayerIds);
   const excludedSet = new Set(excludedLayerIds);
 
+  // Voice dictation. `speech` is null when the browser lacks the Web Speech
+  // API, in which case the mic button is not rendered. Dictation appends to
+  // whatever is already typed, so we snapshot that text when a session starts.
+  const dictationBaseRef = useRef("");
+  const speech = useSpeechInput({
+    onStart: () => {
+      dictationBaseRef.current = inputValue.trim()
+        ? `${inputValue.trim()} `
+        : "";
+    },
+    onResult: (transcript) =>
+      setInputValue(dictationBaseRef.current + transcript),
+  });
+
   // Pills reflect layers/dates active in chat context. Dismissing a pill
   // removes it from context only; the map layer stays visible.
   const datasetPillLayers = layers.filter(
@@ -111,6 +131,7 @@ export default function ChatInput({
   const submitPrompt = async () => {
     if (!inputValue.trim() || isLoading) return;
 
+    speech?.stop();
     const message = inputValue.trim();
     setInputValue("");
     onAfterSend?.();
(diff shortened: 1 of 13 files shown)


kamicut (approved):
Approved! However I couldn't find the feature flag, is it implemented?

01painadam:
> Approved! However I couldn't find the feature flag, is it implemented?

woops, I didnt push - pushing and merging. Cheers for the review

PR #580, opened 2026-07-09
feat(trace-analytics): add taxonomy tagging and analytics drill-down

01painadam:
Adds independent topic/language detection over trace content, a TaxonomySection to surface it, and a way to jump from any analytics aggregate straight into the trace explorer with the underlying rows pre-loaded.


01painadam:
Since this is an isolated internal feature, gonna merge since tests pass

PR #588, opened 2026-07-13
Dashboard chat nudges

01painadam:
Adds an initial state to new conversations in Dashboards, including nudges that surface dashboard capabilities.

TODO
- [x] Start new convo from dashboards
- [x] Open old conversations from dashboards
- [x] define behaviour when moving between map <> dashboard

NOTE: we may need to make a decision on routing: how do we carry conversation thread_ids in the url across app and dashboard pages.

1. We should probably rename existing routes:
  a. `/app` --> `/map`
  b. `/dashboard` --> `/setting` (or profile or similar)
2. Decide on convention for conversation thread_id:
  a. Page `/map/threads/conversation_id` and `/dashboards/dashboard_id/threads/conversation_id`
  b. Arg `/map?thread=conversation_id` and `/dashboards/dashboard_id?thread=conversation_id`
  
For now, this branch defers this decision.

### DIFF
--- a/app/app/(chat)/layout.tsx
+++ b/app/app/(chat)/layout.tsx
@@ -11,6 +11,7 @@ import {
 import { useEffect, useState } from "react";
 
 import ChatPanel from "@/app/ChatPanel";
+import ConversationHistoryDrawer from "@/app/components/ConversationHistoryDrawer";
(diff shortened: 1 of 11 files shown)


guilherme-n (approved):

kamicut (approved):
I fixed a quick malformed JSX component, the rest LGTM

PR #602, opened 2026-07-16
feat(dashboards): add existing analyses to a dashboard from the Analyses pane

kamicut:
## Summary

On a dashboard, opening the **Analyses pane** from the chat input's "Analyses" button now lets you add an existing insight (curated/verified, or one you generated earlier) to the grid. It's added at the next free slot as an insight block and persists across reloads. The AI-generated list is scoped to the dashboard's area by default, with a "This area only" toggle to broaden it.

Most of the machinery already existed on `develop` — this is wiring, not green-field: the pane (`InsightsPanel`), the malformed-record-tolerant `GET /api/insights` gateway, and the add-widget backend (`POST /api/dashboards/:id/widgets`, which appends + persists) were all in place.

## What changed

- **`feat(catalog-card)`** — optional `toggleLabel` / `toggleAriaLabel` / `toggleDisabled` props on `CatalogCard`, defaulting to today's show-on-map behaviour (datasets/areas panels unaffected).
- **`refactor(dashboards)`** — extract the add/remove logic from `AddToDashboardToggle` into a reusable `useAddInsightToDashboard` hook, plus `useCurrentDashboardArea`. The chat-side toggle now just renders off the hook (no behaviour change).
- **`feat(insights)`** — on a dashboard, each pane card's footer toggle becomes add/remove-to-grid instead of the inert show-on-map. Gated to real backend insight ids, so verified fixtures / unsaved in-session analyses render the toggle disabled. The AI list scopes to the dashboard AOI via the `aoi_source`+`aoi_id` filter (both sent only together); the read path (`InsightQuery`) caches thread and area scopes separately.
- **`feat(dashboards)`** — mount `InsightsPanel` on the dashboard route (app layer composes both features; the dashboards feature never imports insights-history). Expose the chat input's "Analyses" button on the dashboard-detail route (it was gated to `/app`), keeping the map-only catalog/area pickers hidden. Compact chat shifts aside for the pane there.

## Design notes

- **Filtering is server-side (API), by necessity.** The `InsightResponse` payload carries no AOI (only `statistics_ids`), so client-side filtering is impossible; the backend matches the `(source, src_id)` pair against the statistics rows. Single-area today matches both contracts (dashboard create is `aois.min(1).max(1)`, and the endpoint takes one pair); multi-area later is a localized change in `useCurrentDashboardArea` + `InsightQuery`, not a re-architecture. See the comment on `useCurrentDashboardArea`.
- **FSD** — the pane is mounted at the app-route layer so the `dashboards` feature never imports `insights-history`; the two hooks are deep-imported into the pane, matching the repo's existing `dashboards/ui/*` deep-import convention (avoids a cycle + bundle bloat; `sideEffects` is unset).

## Testing

- `typecheck`, full suite (594), `lint`, `format:check`, and a production `build` all green.
- New/updated unit tests: `useAddInsightToDashboard` (add/remove/gating), the `aoi_source`+`aoi_id` gateway querystring, `useUserInsights` scope plumbing, and `isDashboardDetailRoute`.
- **Not yet browser-verified:** the pane's visual layout on the dashboard (compact vs full-size chat) — the one spot worth a manual pass.

## Follow-ups (not in scope)

- Default the pane's filter to "AI generated" on a dashboard (currently "In this conversation", which can be empty). Deferred due to a `viewContext` timing nuance.
- Search / quick filters (separate story).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/features/dashboards/ui/useAddInsightToDashboard.ts
+++ b/src/features/dashboards/ui/useAddInsightToDashboard.ts
@@ -1,0 +1,84 @@
+"use client";
+
+import { toaster } from "@/app/components/ui/toaster";
+import useAuthStore from "@/app/store/authStore";
+import useViewContextStore from "@/app/store/viewContextStore";
+import {
+  useAddInsightWidget,
+  useDashboard,
+  useDeleteWidget,
+} from "./dashboardQueries";
+
+export interface AddInsightToDashboard {
+  /** True while the viewer is on a dashboard page (add/remove semantics apply). */
+  active: boolean;
+  /**
+   * True when this insight can actually be added: a dashboard the viewer owns,
+   * with a real persisted insight id. False for other people's dashboards and
+   * for insights that have no backend id (verified fixtures, unsaved in-session
+   * analyses).
+   */
+  addable: boolean;
+  /** True when the insight is already a widget on the current dashboard. */
+  added: boolean;
+  pending: boolean;
+  /** Adds the insight if absent, removes it if present. No-op unless `addable`. */
+  toggle: () => void;
+}
+
+/**
+ * Shared "add this analysis to the current dashboard" behaviour, driven by the
+ * ambient view context (no chat round trip — a plain REST widget add/remove
+ * with the analysis's persisted insight id). Backs both the chat-side
+ * `AddToDashboardToggle` and the Analyses pane's per-card add control. All
+ * charts of one analysis share the insight, so one toggle adds/removes the
+ * whole analysis.
+ */
+export function useAddInsightToDashboard(
+  insightId?: string
+): AddInsightToDashboard {
+  const viewContext = useViewContextStore((s) => s.viewContext);
+  const dashboardId =
+    viewContext?.page === "dashboard" ? viewContext.dashboard_id : "";
+  const userId = useAuthStore((s) => s.userId);
+  // Disabled (and dataless) when dashboardId is "" — off dashboard surfaces the
+  // hook never fetches.
+  const { data: dashboard } = useDashboard(dashboardId);
+  const addWidget = useAddInsightWidget(dashboardId);
+  const deleteWidget = useDeleteWidget(dashboardId);
+
+  const active = dashboardId.length > 0;
+  const isOwner = !!userId && !!dashboard && userId === dashboard.user_id;
+  const addable = active && isOwner && !!insightId;
+  const existing = insightId
+    ? dashboard?.widgets.find((w) => w.insight_id === insightId)
+    : undefined;
+  const pending = addWidget.isPending || deleteWidget.isPending;
+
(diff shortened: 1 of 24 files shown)


src/features/dashboards/ui/useAddInsightToDashboard.ts
```diff
@@ -0,0 +1,84 @@
...
+  insightId?: string
+): AddInsightToDashboard {
+  const viewContext = useViewContextStore((s) => s.viewContext);
+  const dashboardId =
+    viewContext?.page === "dashboard" ? viewContext.dashboard_id : "";
+  const userId = useAuthStore((s) => s.userId);
+  // Disabled (and dataless) when dashboardId is "" — off dashboard surfaces the
+  // hook never fetches.
+  const { data: dashboard } = useDashboard(dashboardId);
+  const addWidget = useAddInsightWidget(dashboardId);
```
01painadam, line 47: useAddInsightWidget is deliberately not optimistic (the POST response lacks insight expansion), and its onSettled fires invalidateQueries without returning the promise. So pending (mutation isPending) goes false as soon as the POST resolves, while added only flips when the detail refetch lands. In that window the switch shows the stale state and is enabled again, meaning a second click re-POSTs the same insight and you get a duplicate widget. The delete path is optimistic, so it's the add path that's exposed. Narrow window, but the pane's card switch invites faster clicking than the old chat button did.

Consider returning the invalidateQueries promise from useAddInsightWidget's onSettled, or treat added || mutation.isSuccess as on.

src/features/insights-history/ui/insights-panel.tsx
```diff
@@ -268,13 +317,26 @@ function InsightsPanelHeader({ onClose }: { onClose: () => void }) {
...
+  areaScoped: boolean;
+}) {
   const currentThreadId = useChatStore((s) => s.currentThreadId);
   const liveWidgets = useInsightStore(useShallow((s) => s.insights));
-  // Both queries are cached separately by thread id; when there is no thread the
+  const dashboardArea = useCurrentDashboardArea();
+  // On a dashboard, scope the AI list to its area (both aoi params travel
+  // together). Off a dashboard, or when the "This area" toggle is off, the query
+  // is unscoped. Each scope caches under its own key, so toggling is instant.
+  const aiScope = areaScoped && dashboardArea ? dashboardArea : undefined;
```
01painadam, line 345: when the dashboard AOI is unavailable (detail loading or errored), the AI list quietly falls back to unscoped while the switch stays checked. Gate the list (or disable the switch) on dashboardArea being resolved, so the label never overstates the scope.

01painadam (approved):
Two minor things to address, but otherwise looks good!

01painadam (approved):

PR #604, opened 2026-07-17
Dashboard widget layout: full-width chart content + content-height packing

01painadam:
## Summary

Three related fixes to how dashboard widgets use space, addressing the "stretched widgets / empty voids" and "full-width chart looks broken" tickets:

### 1. Map widgets default to full width (`c061bba`)
`mapWidgetSize()` reads `config.size` with a `double` default, so maps span both columns unless explicitly shrunk.

### 2. Full-width chart widgets use their space deliberately (`1b5fbc8`)
Previously a full-width donut kept its small fixed radius while its legend (recharts `align="right"`, `width: 42%` of the container) pinned to the far card edge — reading as a rendering bug.

- The card's width state now threads down: `DashboardWidgetCard` → `WidgetMessage` (new `fullWidth` prop) → `ChartWidget`
- At full width, a **pie** caps its chart+legend block at the single-column content width (~540px) and centers it — legend adjacent to the chart, symmetric margins. Cartesian charts already stretch and are unchanged
- Half-width and chat rendering are untouched; fullscreen keeps its own `expanded` sizing
- `/chart-debug` gains a **"Full-width card"** toggle to QA both states side by side

### 3. Widgets pack by content height — no more voids (`2bfc1ba`)
Cards had `h="100%"` inside CSS grid rows, so every card stretched to its row's tallest neighbour and short widgets left large empty blocks.

- **New `packCells()`** (`src/features/dashboards/lib/packing.ts`): full-width cells get their own row; runs of consecutive half-width cells deal into two tightly-stacked columns (evens left, odds right). Adjacent halves still pair side by side for comparison
- Packing is **content-blind** — persisted order + sizes only, never rendered heights — so the layout is stable across loads and never reshuffles as charts animate
- Drag-and-drop reorder is untouched (same handlers, same flat-index mapping)
- **Mobile**: column wrappers flatten away (`display: contents`) and per-card `order` restores the flat arrangement — pure CSS, no breakpoint JS
- Text notes (the only unbounded content — tables paginate, maps/charts are fixed) cap at 520px with internal scroll and keep an 80px floor

**Trade-off flagged for design:** order beats height-balance — if widget 1 is very tall and 2–4 short, widget 3 packs under 1 rather than under 2. Height-greedy packing would fill space slightly better but re-flows as content loads and breaks the adjacent-pair mental model.

## Test plan

- [x] `format:check` / `lint` / `typecheck` / `pnpm test` all pass (597 tests, 9 new for `packCells`)
- [x] Verified live against a mock backend with a mixed-height fixture (short donut + 28-row table + bar + note + map + long note): bar packs up under the donut beside the tall table; no voids
- [x] Width toggle both directions re-packs cleanly; full-width note spans its own row
- [x] Refresh reproduces the exact layout (persisted `config.size` + deterministic packing)
- [x] Long note scrolls internally at its cap
- [x] Phone viewport (400px): single column in exact arrangement order
- [x] Full-width donut on `/chart-debug`: capped + centered, legend adjacent; single-width unchanged
- [x] QA on staging with real dashboards (drag reorder across packed columns, multi-chart insights)
- [x] Design sign-off on the 540px pie cap and 520px note cap (constants, trivial to retune)


kamicut (approved):
Two layout bugs worth investigating here or as a separate PR:
1. In a narrow screen, visual order is different from "reader-mode" order, which breaks accessibility. To test: make the screen single column and try tabbing through 5 cards, the tab will skip around because the reader-mode order is interleaved rather than linear.
2. When switching a single column chart from a chart to a table, a re-ordering switches the chart back to "chart" because the state doesn't survive remounts.

PR #606, opened 2026-07-17
feat(dashboards): nest areas and improve search UX on the new-dashboa…

01painadam:
Adds a few new QOL features on the /dashboards page:
- Compact nested areas on area search
- Loading state when searching
- Actively searches as soon as at least 3 characters are entered: no enter is required

### DIFF
--- a/src/features/dashboards/ui/NewDashboardScreen.tsx
+++ b/src/features/dashboards/ui/NewDashboardScreen.tsx
@@ -1,26 +1,9 @@
 "use client";
 
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { useRouter } from "next/navigation";
-import {
-  Box,
-  Button,
-  Flex,
-  IconButton,
-  Input,
-  Menu,
-  Portal,
-  Spinner,
-  Table,
-  Text,
-} from "@chakra-ui/react";
-import {
-  CaretRightIcon,
-  DotsThreeVerticalIcon,
-  MagnifyingGlassIcon,
-  PolygonIcon,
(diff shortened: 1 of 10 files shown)


src/features/dashboards/ui/NewDashboardScreen.tsx
```diff
@@ -526,6 +339,21 @@ export function NewDashboardScreen() {
...
+      {trimmedSearch && !showSkeletons && rows.length > 0 && (
+        <Text
+          mt="10px"
+          fontFamily="mono"
+          fontSize="11px"
+          letterSpacing="0.4px"
+          color="#9AA3B2"
+          textTransform="uppercase"
+        >
+          {tree.length}
```
kamicut: Should this count only the root? (i.e searching for Brazil yields '14 results' instead of 63). This is also not a true count since the table is paginated so we're not sure what the real number is going to be. We would need that returned in the API.

src/features/dashboards/ui/NewDashboardScreen.tsx
```diff
@@ -526,6 +339,21 @@ export function NewDashboardScreen() {
...
+          {hasNextPage ? "+" : ""} areas match &ldquo;{trimmedSearch}&rdquo;
```
kamicut: Consider dealing with plural and singular ({rows.length === 1 ? "area" : "areas"})

src/features/dashboards/ui/NewDashboardScreen.tsx
01painadam: tch, true - didnt consider that. I'll remove

PR #609, opened 2026-07-21
feat(dashboards): adaptive header — pin and condense on scroll

01painadam:
## Summary

Implements the adaptive dashboard header (Figma: *Dashboard default_Adaptative header banner*), plus a fix for the widget grid overflowing when the full-size chat narrows the content area.

Screenshot 2026-07-23 at 13 45 54

### Adaptive header (`51ae4bc`)

On long, widget-heavy dashboards the header (breadcrumb, title, "Updated…" timestamp, Export/Share) scrolled away with the page. Now:

- At the top of the page the header renders in full, unchanged.
- Once the header is completely hidden behind the 40px global nav, a **fixed condensed header** fades in below the nav: the breadcrumb on a page-background strip, and a 104px white bar with the card's 2px blue top accent, rounded top corners, and an elevation shadow — widgets scroll underneath it.
- Scrolling back to the top restores the full header. Short dashboards that don't scroll never pin (IntersectionObserver on the in-page header, so there's no scroll-listener work per frame).
- The pinned bar reuses `DashboardHeader` (new `condensed` prop: single-line title with ellipsis, `h2` instead of a second `h1`), so rename/Export/Share behave identically in both states by construction.
- Whichever copy of the header is off-screen is removed from the tab order and accessibility tree (`aria-hidden` + `inert`), so assistive tech and keyboard users always see exactly one live set of controls.

### Widget grid overflow fix (`b859942`)

The grid chose 1 vs 2 columns from the `lg` **viewport** breakpoint, but the full-size chat is a fixed overlay that removes ~430px of content width without changing the viewport — so at common window sizes the grid kept two columns the cards couldn't fit into, and the right column overflowed off-screen. The grid wrapper is now a CSS inline-size container; the column count and the "double" span switch on the grid's own width (`@container (min-width: 700px)`), with `minW={0}` on items as a backstop so tracks can never force the page wider.

## Test plan

- [x] 14 new unit/component tests: pin/unpin logic (incl. below-viewport and no-IntersectionObserver branches), pinned-bar content and false-door toast parity, one-live-copy a11y invariant at page level
- [x] `pnpm typecheck` / `pnpm lint` / `pnpm format:check` / `pnpm build` all pass
- [x] Browser-verified (Playwright + mocked backend): pin → condense → unpin cycle; Export/Share from the pinned bar; 640px viewport title truncation; grid at 1150/1280/1440px in compact and full-size chat with zero horizontal overflow; "double" widgets span correctly in both column modes
- [ ] QA on staging with real dashboards (many widgets, long dashboard names, rename from the pinned bar)
- [ ] Cross-check condensed-bar spacing against Figma in Dev Mode

Note: the 4 failing `NewDashboardScreen` tests fail identically on clean `develop` (pre-existing, unrelated to this PR).


src/features/dashboards/ui/DashboardWidgetsGrid.tsx
kamicut, line 36: nit: this JSDoc belongs to `GridCell` so should be moved to line 50

kamicut (approved):

PR #628, opened 2026-08-06
PZB-1158: Curated insights carry a title

guilherme-n:
## Pull request type

- [x] Feature

Curated (verified) insight fixtures have no title of their own. The card that displays them derives its title from the first chart's title field (e.g. "Annual tree cover loss"), which doesn't identify which area the analysis covers. This makes curated insights hard to tell apart on a dashboard or in the insights list without opening the chart.

Issue Number: PZB-1158

## What is the new behavior?
- Added generateInsightTitle({ datasetName, locationName, areaLabel }), producing a title in the form "{dataset} in {location}", falling back to the area's own label when a resolved location name is unavailable.
- InsightRecord gained an optional title field: only curated insights set it. AI-generated insights are untouched and keep deriving their title from the first chart, unchanged.
- Added resolveInsightTitle(record, chartTitle) and wired it into insights-panel.tsx so the record's generated title (when present) wins over the per-chart title, keeping the same title wherever the insight is rendered.
- The "Tree cover loss in Brazil" curated fixture now carries a generated title. The KBA fixture is left without one on purpose - it breaks protection down by region rather than covering a single location, so the "{dataset} in {location}" template doesn't fit it.
- Unit tests for title generation across dataset/area combinations (including the missing-location fallback) and for the resolve/override behavior.

## Demo
Screenshot 2026-08-05 at 20 56 18

_______________

Screenshot 2026-08-06 at 15 06 31

### DIFF
--- a/src/features/insights-history/ui/insights-panel.tsx
+++ b/src/features/insights-history/ui/insights-panel.tsx
@@ -74,10 +74,6 @@ const INSIGHT_SELECTED_BG = "rgba(0, 73, 170, 0.06)";
 
 type InsightFilter = "conversation" | "verified" | "ai";
 
-// The "Curated" (verified) filter is hidden for now: product is rearchitecting
-// curated analyses, so the hand-written fixtures shouldn't be surfaced. The
-// filter branch and `verifiedInsights` are left in place so putting the chip
-// back is a one-line change once the real source exists.
 const INSIGHT_FILTERS: { id: InsightFilter; label: string }[] = [
   { id: "conversation", label: "In this conversation" },
   { id: "ai", label: "AI generated" },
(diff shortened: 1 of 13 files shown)


01painadam (approved):
LGTM let's also show the Curate filter on the analysis panel

src/features/insights-history/ui/insights-panel.tsx
01painadam: Small nit - lets remove this

01painadam: Should this be:

```
widget: {
      ...widget,
      title: resolveInsightTitle(record, widget.title),
      curated,
    },
```

I think you had this previously

01painadam (changes requested):
I think a small change is needed - note that I added a small commit to catch an edge case.

01painadam (approved):

PR #630, opened 2026-08-08
create or open dashboard from map aoi

guilherme-n:
## Pull request type
- [x] Feature

Selecting an administrative area on the map (with a dataset active) surfaces two in-chat nudges: **Generate {dataset} Insight for {area}**, which sends a prompt to the agent, and **View {dataset} Analysis for {area}**, which runs the analytics API directly. Neither leads to a dashboard.

A dashboard can only come into existence two ways:

- **Manually**, from `/dashboards` → *New Dashboard*, picking an area out of a searchable table, completely disconnected from whatever the user is looking at on the map.
- **Via the agent**, when it decides to call `create_dashboard` mid-conversation (PZB-1062 / #608), which surfaces a DASHBOARD navigation card in the thread.

So there is no way to turn the area you are already looking at into a dashboard in one step. There is also no on-map affordance for the selected area at all: no chip, no actions menu. Analysing it, saving it or clearing it live in unrelated corners of the UI. And nothing checks whether the selected area *already* has a dashboard, so going through the flow twice silently produces two dashboards for the same AOI.

## What is the new behavior?

Two manual entry points, both pure REST, with no agent round trip and nothing touching the chat stream pipeline.

**1. In-chat nudge**, a third card alongside the two existing ones, reading `Create Dashboard for {area}`. It surfaces on the same gate as its siblings (area selected plus dataset active), and additionally requires `?ff=dashboard` and a resolvable `src_id`/`subtype`. Without those, `POST /api/dashboards` could only fail, so the nudge stays hidden.

**2. On-map AOI chip.** The selected area now renders as a chip over the map with its name, an `✕`, and a `⋮` menu:

- **ANALYSIS**: Generate Insights, View Analysis
- **MANAGE**: Create/Open Dashboard, Save Area, Remove from map

Every item delegates to behaviour that already existed. Items whose preconditions aren't met are hidden rather than disabled, the same choice `AddToDashboardToggle` makes.

Both entry points share one hook, `useCreateDashboardForArea`:

- If a dashboard already covers the AOI, the label becomes `Open {area} Dashboard` and the click navigates straight to `/dashboards/:id?ff=dashboard`. Matching is `source` plus `src_id`, with a case-insensitive `name` fallback and `updated_at` breaking ties. An id match always beats a name match.
- Otherwise it creates one. The payload carries the AOI and **no name**, so the backend titles the dashboard by its area. The DASHBOARD card goes up in chat immediately, then the non-generative analysis for the active dataset runs and is attached as the dashboard's first widget. The card's subtitle ticks from 0 to 1 as it lands.
- Ordering is deliberate. The analysis is a poll-to-completion job (roughly 30 to 60 seconds), so blocking the card on it would leave a dead button. Failure is split the same way: a failed create means nothing was persisted (toast, no card), while a failed analysis leaves a usable dashboard already on screen (warning only).
- After creating, the nudge relabels *itself* to `Open …` off the invalidated dashboards query, so there is no `accepted` flag on the message.

The chat card, the `dashboard-card` message type and the assistant copy are reused from #608. That PR's inline wording is extracted into shared helpers, so a dashboard created by hand reads identically to one the agent created.

### Not covered

`analysisSelection` is only ever set by a manual click on a GADM area (`VectorAreasLayer`), so neither entry point appears for an AOI the *agent* picked, or after a thread reload. That predates this PR (the two existing nudges have the same limitation) and is left as-is deliberately, since changing it would also change when Generate Insights and View Analysis appear.

## Demo
Screenshot 2026-08-10 at 15 37 52

______

Screenshot 2026-08-07 at 21 49 41


01painadam (approved):
Nice - looks good to me!

01painadam (approved):
Great stuff! Merging.

PR #631, opened 2026-08-09
PZB-1169: feat(dashboards): group insights into modules with per-chart Customize menu

kamicut:
## What

Dashboards now render each insight as a **grouped module** instead of flattened per-chart cards, per @faustoperez's prototype (https://gnw-dashboard-insight-groupings.vercel.app):

- **Insight modules on the grid** — each `widget_type: "insight"` widget renders full-width with a header row (drag handle, collapse chevron, title, Customize, remove-with-confirm), the AI-generated summary with its AI-ASSISTED badge (previously only reachable in the chart fullscreen dialog), and the insight's chart cards packed side by side inside the module.
- **Customize menu** — a per-module popover with checkbox rows for the summary and every chart ("Chart · <name>", hidden ones included). Toggles persist immediately to the widget `config` (`summaryHidden`, `chartIds`) through the existing optimistic mutations; the panel stays open for multiple toggles and remaining cards reflow live.
- **Analyses panel regroup** — on a dashboard, the sidebar now lists one card per analysis with a whole-insight "Add to dashboard" switch; per-chart visibility moved to the Customize menu. The map surface keeps its one-card-per-chart behaviour. Detail view pages within the selected analysis's charts on dashboards.

## Behaviour changes

- Hiding an insight's **last** chart no longer deletes the widget — the module survives with `chartIds: []` (its summary can still render, and Customize can re-show charts).
- Reordering is unchanged in semantics (always widget-level) but simpler: dragging a module header or any chart handle moves the whole module.
- Module title falls back `config.title` → first chart title → "Analysis" (backend sends no insight title; noted as a follow-up).

## Out of scope (agreed during planning)

- Map widgets joining insight modules (open design question — maps/notes render standalone, unchanged).
- The prototype's comment/annotation icon (separate feature).
- Collapse-state persistence (local-only for now).

## Testing

- `pnpm test` — 791 tests / 95 files green, including new coverage for the visibility helpers (`withChartHidden` empty-set contract, `summaryHidden`, `moduleTitle`, `insightModule` view-model), `DashboardInsightModule` render matrix, the Customize menu, grid grouping, and the panel's grouping lib.
- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm test:arch`, `pnpm build` all pass.
- Manual QA still needed on a real owned dashboard (auth required): module rendering/reorder/persistence, Customize toggles across reload, non-owner/public view, and the Analyses panel on both surfaces.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/features/dashboards/ui/DashboardInsightModule.tsx
```diff
@@ -0,0 +1,352 @@
...
+      <Box
+        key={card.id ?? order}
+        minW={0}
+        css={{ order, [TWO_COLUMN_QUERY]: { order: 0 } }}
+      >
+        <DashboardWidgetCard
+          title={card.title}
+          card={card}
+          placeholder={null}
+          chartCount={vm.cards.length}
```
01painadam: This passes the count of shown cards, which DashboardWidgetCard uses to choose its remove-confirmation copy (chartCount > 1 gives "Remove chart?", anything else gives "Remove widget? / The widget will be removed from this dashboard.")

Since withChartHidden no longer deletes the widget, that second branch now describes something that doesn't happen. Confirmed by rendering: a three-chart insight with chartIds: ["c-1"] opens a dialog titled "Remove widget?" whose body reads "The widget will be removed from this dashboard" and clicking through only hides the chart, leaving the module in place.

Simplest fix is chartCount={vm.allCharts.length} so the copy tracks the roster rather than the visible subset; cleaner would be an explicit copy prop, since inside a module the card's X never removes a widget regardless of count.

src/features/insights-history/ui/insights-panel.tsx
```diff
@@ -409,6 +471,49 @@ function EmptyState({ filter }: { filter: InsightFilter }) {
...
+}) {
+  const insight = useAddInsightToDashboard(group.addableInsightId);
+  const title = group.title;
+
+  return (
+    <Box w={`${CATALOG_CARD_WIDTH_PX}px`} maxW="100%" flexShrink={0}>
+      <CatalogCard
+        thumbnail={<InsightThumbnail type={group.widgets[0]?.type ?? "bar"} />}
+        typeLabel="ANALYSIS"
+        typeLabelColor={INSIGHT_LABEL_COLOR}
```
01painadam, line 496: Flicking this toggle off deletes the entire widget via useAddInsightToDashboard, discarding the module's per-chart sizes, renames, chart visibility and summary setting with no confirmation and no undo — while removing the very same module from the grid requires confirming "Remove analysis?" first. The old per-chart toggle also removed without confirming, but it only cost one chart's visibility flag; grouping raised the blast radius to the whole module's customisation. Consider matching the grid's confirmation when the widget carries a non-empty config, or at least a toast with an undo.

src/features/insights-history/lib/insight-groups.ts
```diff
@@ -0,0 +1,97 @@
+import {
+  chartsToWidgets,
+  resolveInsightTitle,
```
01painadam, line 4: Two independent title rules now exist for one analysis: this uses resolveInsightTitle over the first chart in the gateway's array order, while moduleTitle in dashboards/lib/widgets.ts uses config.title then the first chart in position order then "Analysis". They coincide only because RestInsightsGateway.toInsightRecord happens to set position: i from the array index, and neither function references the other. If the dashboards API ever returns charts in a different order from its position field, the same analysis will be headed one way in the panel and another on the grid. A shared title resolver (or at minimum a comment in each pointing at the other) would make the coupling visible.

01painadam (changes requested):
Small changes requested otherwise good.

PR #645, opened 2026-08-26
[PoC][PZB-1185] Can mixed-resolution LGMS layers roll up into a single map layer group?

willian-viana:
The MVP map needs a map layer for LULUCF and Agri for the same data the dashboard charts show, sharing definitions, sign convention and units so on-map sums reconcile with on-chart numbers.

Additionally, we would like to add  a parent "Full LGMS" layer (nice-to-have, but worth understanding). The two child layers are at very different resolutions (30 m LULUCF and 10 km Agri) and we do not know whether that rollup should be rendered and aggregated without misleading users, and we cannot scope the implementation Stories until we do.

## Background: the layers in scope

The spike is investigating the rollup against this intended MVP layer set. This is context, not a build instruction.

- **(MVP) Agri 2020**: 10 km, 2020 only. MVP is the single combined layer; Crop plus Livestock split is a nice to have

- **(MVP) Net Annual LULUCF**: one combined LULUCF layer at 30 m; net flux only; annual average (2016 to 2024)

- **(P1) Full LGMS**: first hierarchy level, the sum of the two layers above

The child layers are different resolutions, which is the crux of the problem.

## Questions to Answer

How we will define tiles for the LULUCF (2016-2024 annual avg) and Agri (2020 fixed) layers

- Can a 30 m layer and a 10 km layer be combined into a single displayable "Full LGMS" parent layer, and if so at what target resolution?
- What does the combination do to the numbers? Does a resampled or aggregated parent layer still reconcile with the dashboard chart figures for the same area, within an acceptable tolerance?
- How should the resolution mismatch be surfaced to the user so the parent layer is not read as 30 m precision when half its signal is 10 km?
- What are the temporal implications of summing an annual average (2016 to 2024) with a single year (2020), and is the parent layer defensible at all given that mismatch?
Is the tiling and serving cost for the 30 m layer acceptable at MVP, or does it force a compromise on resolution or extent?

## Data Infra Perspectives

- How we will define tiles for the LULUCF (2016-2024 annual avg) and Agri (2020 fixed) layers?

## WIP
Update: we have path forward to add multilayer support to zeno.

> Can a 30 m layer and a 10 km layer be combined into a single displayable "Full LGMS" parent layer, and if so at what target resolution?

Zonal stats are based on upsampled Agri data, so we may want to do that here for consistency. It’s technically feasible to generate a COG for this.

> What does the combination do to the numbers? Does a resampled or aggregated parent layer still reconcile with the dashboard chart figures for the same area, within an acceptable tolerance?

Chart figures are based on upsampled Agri data at 30m, so should be consistent.

> What are the temporal implications of summing an annual average (2016 to 2024) with a single year (2020), and is the parent layer defensible at all given that mismatch?

## Product/Research question?

> Is the tiling and serving cost for the 30 m layer acceptable at MVP, or does it force a compromise on resolution or extent?

Unlikely to be an issue given current costs, but getting definitive numbers is WIP.
update: confirmed COG creation cost for agri and LULUCF was <$50

### Answered
- Q3.1 (how to surface the resolution mismatch to the user) — caption + warning banner give a working UI proposal.

### Partially answered
- Q3 (does the rollup reconcile with dashboard figures within tolerance) — builds a mismatch-flagging mechanism, but it checks internal self-consistency within one payload, not against the actual dashboard chart figures.
- Q4 (temporal implications of summing an annual average with a single year) — visually surfaced in the caption, but "is the parent defensible" is left open, as the ticket itself marks it a product/research question.

### Not answered
- Q1 (how tiles will be defined for LULUCF and Agri) — infra/tiling concern, out of scope for a chart PoC.
- Q2 (can 30m + 10km combine into a displayable map layer, at what resolution) — this PoC renders a chart, not a map layer, so it doesn't touch this.
- Q5 (tiling/serving cost for the 30m layer at MVP) — infra/cost concern, not something frontend code can determine.

## Artifact
https://claude.ai/code/artifact/083c5357-3760-484d-bf3f-528b11d9b1b8

### DIFF
--- a/app/chart-debug/ChartDebugPanel.tsx
+++ b/app/chart-debug/ChartDebugPanel.tsx
@@ -21,7 +21,28 @@ import CHART_COLOR_MAPPING, {
   DATASET_DIVERGENT_COLORS,
 } from "@/app/config/chartColorMappings";
 import getChartColors from "@/app/utils/ChartColors";
-import { GHG_FLUX_TREE_DUMMY_WIDGET } from "@/src/features/ghg-flux-tree";
+import {
+  GHG_FLUX_TREE_DUMMY_RESULT,
+  GHG_FLUX_TREE_DUMMY_WIDGET,
+} from "@/src/features/ghg-flux-tree";
+
+// PoC fixture (PZB-1185): the same tree, but "all_land" no longer equals the
+// sum of its "land_use" + "agriculture" children — standing in for a
+// resampled/independently-aggregated "Full LGMS" rollup that has drifted out
+// of reconciliation. Exercises GhgFluxTreeBody's reconciliation-warning
+// banner, which the clean fixture above never triggers.
+const GHG_FLUX_TREE_MISMATCH_WIDGET: InsightWidget = {
+  ...GHG_FLUX_TREE_DUMMY_WIDGET,
+  id: "ghg-flux-tree-mismatch-dummy",
+  title: "Net GHG flux (annual average) — reconciliation mismatch",
+  data: (
+    GHG_FLUX_TREE_DUMMY_RESULT.charts[0].data as Record<string, unknown>[]
+  ).map((row) =>
+    row.id === "all_land"
+      ? { ...row, avg_emissions: 1750, avg_removals: -700 }
+      : row
+  ),
+};
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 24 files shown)


willian-viana:
Handling it in #650

PR #646, opened 2026-08-27
Add IFL to existing dataset drawer UI

guilherme-n:
## Pull request type
- [x] Feature

## What is the new behavior?
- Adds Intact Forest Landscapes (IFL) as a standalone card in the Data Catalog and the mobile layer picker, badged VIEW ONLY since it has no analytics endpoint. Name, colors and description are reused from the existing intact_forest context-layer metadata, so the standalone layer and the sub-layer already rendered under Tree Cover Loss stay identical.
- The card is gated behind a new ?ff=ifl feature flag for researcher review, per the ticket. Hidden from both browse surfaces until the flag is set; a layer already on the map keeps rendering regardless (so dropping the flag mid-session doesn't break an active layer).
- Toggling the card on adds the IFL raster layer to the map and renders its legend (title, 4-color swatch list, opacity/visibility controls) exactly like any other dataset card, reusing the existing layer-manager and legend pipeline, no new UI primitives.
- Since IFL is not analysable, the analyse-nudge, view-analysis-nudge, and create-dashboard-nudge triggers now skip view-only datasets when picking the "active dataset" for their CTA, falling through to the first analysable layer if one is co-active. Without this, selecting an area with only IFL on the map would nudge toward an analysis that can't run.
- Adds enabledFlags() / useEnabledFlags() to the shared feature-flag utility, for gating a whole list against multiple flags in one call instead of one useFeatureFlag per flag.

## Demo
Screenshot 2026-08-26 at 22 50 38

### DIFF
--- a/app/components/CatalogPanel.tsx
+++ b/app/components/CatalogPanel.tsx
@@ -19,6 +19,7 @@ import {
   EyeIcon,
   EyeSlashIcon,
   StackPlusIcon,
+  StackSimpleIcon,
   XIcon,
 } from "@phosphor-icons/react";
 import { useShallow } from "zustand/react/shallow";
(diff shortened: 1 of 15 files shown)


01painadam (approved):
Nice! LGTM.

I added a small commit to improve the fallback icon to be a bit prettier. It was an existing issue that was bugging me and it fits nicely in this PR.

PR #648, opened 2026-08-27
Migrate from Next.js to Vite and React Router

kamicut:
## What this PR does

This is the long running branch for moving the app off Next.js and onto Vite with React Router v7 (library mode). The work lands as small commits. Each commit builds, typechecks, lints, and passes tests on its own, so you can review and run the app at any commit.

Background: the app is already a client side SPA. There are no Next API routes, no middleware, no server sessions, and no RSC data fetching. All data goes straight from the browser to the FastAPI backend. The only server side pieces are the HTML shell and one server action, and this PR removes both dependencies on Next step by step.

### Phase 1: remove Next-isms while still on Next (done)

Every commit here keeps the app fully working on Next and deployable as before.

- [x] 1. Remove unused deps (next-themes, cookie) and the dead svgr turbopack rule
- [x] 2. Replace next/image with Chakra Image in the 4 files that used it
- [x] 3. Load IBM Plex fonts from @fontsource instead of next/font, theme tokens point at the family names directly
- [x] 4. Load GTM with a plain script tag instead of next/script
- [x] 5. Fetch provenance data client side and delete the only server action. CORS spot check passed against the staging API; full check with a live insight happens in phase 3 QA
- [x] 6. Add an internal router wrapper (app/lib/router.ts) that re-exports next/link and next/navigation
- [x] 7. Migrate all consumers (37 files) to the router wrapper, update 5 test mocks, add the no-restricted-imports lint rule

### Phase 2: cutover to Vite

These commits are one atomic unit. The app runs on Vite from the flip commit onward.

- [ ] 8. Add vite entry (index.html, main.tsx) and the React Router route tree. Pure addition, Next still builds
- [ ] 9. Switch build tooling from Next to Vite (vite.config.ts, tsconfig, eslint, scripts)
- [ ] 10. Remove next and eslint-config-next from dependencies
- [ ] 11. Update amplify.yml and CI for the Vite build output, add SPA rewrites, remove .vercel

### Phase 3: QA and cleanup

- [ ] 12. Remove Suspense boundaries that only existed for the Next useSearchParams rule
- [ ] 13. QA checklist on vite preview: landing page, auth flow, chat streaming, thread deep links, map and AOI, dashboards, trace analytics, debug pages, 404, analytics scripts

## Notes for reviewers

- The riskiest change was commit 5. It moves an external fetch from the server to the browser. The staging data API already answers with access-control-allow-origin for app origins, so the browser call is allowed. End to end confirmation with a live insight is part of phase 3 QA.
- Commit 7 is large but every hunk is the same import path change. Only router.push and router.replace are used anywhere, so the future React Router shim in the wrapper stays small.
- The lint rule moved from commit 6 to commit 7 so that every commit passes repo wide lint.
- Dynamic routes (/app/threads/:id, /dashboards/:id) mean the app can never use static export on Next. That is why there is no SSG step. QA happens on the Vite preview build.


app/lib/router.ts
```diff
@@ -0,0 +1,12 @@
+// Single indirection point for routing APIs. The Vite + React Router
+// cutover reimplements this file and leaves consumers untouched, so
+// nothing outside it may import next/link or next/navigation directly
+// (enforced by no-restricted-imports in eslint.config.mjs).
+export { default as Link } from "next/link";
+export {
```
willian-viana: Nice! I'd suggest even for a temporary abstraction, to move it into the FSD structure. `src/app` is the proper place to keep it.

willian-viana (approved):
:shipit:

